PR #1, opened 2020-09-10
Give it a name

tomusdrw:
Alternative name:
BEEP = BridgeEfficiencyEnablingProtocol

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,20 +22,20 @@ This means that we could verify finalilty interactively like this:
 ### General Notes
 Every time we say “Merkle root/merkelize” we mean `keccak256` ordered merkle trie (`keccak_256_ordered_root` function in Substrate).
 
-### BridgeGadget Substrate Pallet
+### BEEFY Substrate Pallet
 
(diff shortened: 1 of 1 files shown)


svyatonik (approved):
I have no opinion on how to call it :) But imo if it ever ends in Substrate/Polkadot client/runtime code, then either it should be something wide-known, or it should have `grandpa` preffix/suffix in its name. `secondary-grandpa` probably?

P.S.: when you guys decide about its name, I suggest to use that name in other parts of the README as well - i.e. in `BridgeGadget`, `BridgeGadgetToMMR`, `GrandpaBridgeGadget`.

andresilva (approved):
Name LGTM, I guess we can still change it in the near future if someone comes up with anything better.

I'd like the name to not relate to GRANDPA as the protocol is more detached from GRANDPA than it seems from what we've been talking. It's true that we rely on some assumptions from GRANDPA but I don't think those are specific to GRANDPA and would probably have been served by some other finality gadget as well (once we have a clearer view we can specify what these assumptions are).

PR #4, opened 2020-10-16
Initial project setup and skeleton

andresilva:
This PR adds the initial skeleton for the client-side of the BEEFY gadget. There isn't any BEEFY specific logic implemented yet, and only includes the initial setup for:
- listening to finality events;
- gossiping votes;
- accumulating votes for a given round.

Additionally it also adds a folder `node-example` which is the `node-template` code configured with the BEEFY gadget. Everything there is unchanged from node-template except for a change in `service.rs` to start the gadget. I'm unsure whether we should add this here but I think it will be useful for development (e.g. soon enough we'll also need to add the BEEFY and MMR pallets to the runtime).


beefy-gadget/src/lib.rs
```diff
@@ -0,0 +1,392 @@
+use std::collections::BTreeMap;
```
andresilva, line 1: What license do we use for this? I guess GPL-3 since it's client code?

node-example/node/src/service.rs
```diff
@@ -0,0 +1,314 @@
...
+			keystore.clone(),
+			can_author_with,
+		)?;
+
+		// the AURA authoring task is considered essential, i.e. if it
+		// fails we take down the service with it.
+		task_manager.spawn_essential_handle().spawn_blocking("aura", aura);
+
+		// Start the BEEFY bridge gadget.
+		task_manager.spawn_essential_handle().spawn_blocking(
```
andresilva, line 175: This is the only change from the `node-template` so far.

beefy-gadget/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+	Signature: Codec + Debug + PartialEq + std::convert::TryFrom<Vec<u8>>,
+	FinalityNotifications: Stream<Item = FinalityNotification<Block>> + Unpin,
+{
+	fn should_vote_on(&self, number: NumberFor<Block>) -> bool {
+		use sp_runtime::traits::Saturating;
+		use sp_runtime::SaturatedConversion;
+
+		let diff = self.best_finalized_block.saturating_sub(self.best_block_voted_on);
+		let diff = diff.saturated_into::<u32>();
+		let next_power_of_two = 2u32.pow((diff as f64 / 2.0).log2().ceil() as u32);
```
tomusdrw, line 203: Why not `https://doc.rust-lang.org/std/primitive.u32.html#method.next_power_of_two`?
```
(diff / 2).next_power_of_two()
```

beefy-gadget/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+		let next_block_to_vote_on = self.best_block_voted_on + 2.max(next_power_of_two).into();
```
tomusdrw, line 204: The `2` here might be configurable at some point.

node-example/.github/ISSUE_TEMPLATE/ask-a-question.md
```diff
@@ -0,0 +1,13 @@
+---
```
tomusdrw, line 1: Do we need all of this?

node-example/node/src/service.rs
tomusdrw: That's quite some boilerplate then :D

tomusdrw (approved):
lgtm!

node-example/node/src/service.rs
andresilva: Yeah, indeed! And apart from this the whole thing I see changing in the future is:
- Configuring block import pipeline in this file
- Adding some pallets to the runtime

Do you think we should instead create a new repo where we have this example node code? I think it will be useful to have this somewhere.

tomusdrw: No, it's fine. Let's keep it in this repo to make it easier to test.

andresilva:
Addressed comments. Should be good to merge.

beefy-gadget/src/lib.rs
```diff
@@ -177,6 +194,7 @@ where
 		BeefyWorker {
 			local_id,
 			key_store,
+			max_periodicity: 2,
```
tomusdrw, line 197: I think it's rather `min_periodicity`, no?

tomusdrw (approved):

PR #8, opened 2020-10-20
Primitives & Light Client examples

tomusdrw:
PR adds a set of beefy primitives with elaborate hints how the light client should be implemented. The goal for the examples is to serve as a living documentation for BEEFY light client implementation.


tomusdrw:
@andresilva could you take a look? I plan to work on the 2-phase light client samples next, but would prefer to do it in a subsequent PR.

primitives/src/commitment.rs
```diff
@@ -0,0 +1,203 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use crate::ValidatorSetId;
+use core::cmp;
+
+/// A commitment signed by Grandpa validators as part of BEEFY protocol.
+///
+/// The commitment contins a [payload] extracted from the finalized block at height [block_number].
```
andresilva, line 22:
```suggestion
/// The commitment contains a [payload] extracted from the finalized block at height [block_number].
```

primitives/src/commitment.rs
```diff
@@ -0,0 +1,203 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use crate::ValidatorSetId;
+use core::cmp;
+
+/// A commitment signed by Grandpa validators as part of BEEFY protocol.
```
andresilva, line 20:
```suggestion
/// A commitment signed by GRANDPA validators as part of BEEFY protocol.
```

primitives/src/commitment.rs
```diff
@@ -0,0 +1,203 @@
...
+///
+/// The commitment contins a [payload] extracted from the finalized block at height [block_number].
+/// Grandpa validators collect signatures on commitments and a stream of such signed commitments
```
andresilva, line 23:
```suggestion
/// GRANDPA validators collect signatures on commitments and a stream of such signed commitments
```

primitives/src/commitment.rs
```diff
@@ -0,0 +1,203 @@
...
+/// (see [SignedCommitment]) forms the BEEFY protocol.
+#[derive(Debug, PartialEq, Eq, codec::Encode, codec::Decode)]
+pub struct Commitment<TBlockNumber, TPayload> {
+	/// The payload being signed.
+	///
+	/// This should be some form of cummulative representation of the chain (think MMR root hash).
```
andresilva, line 29:
```suggestion
	/// This should be some form of cumulative representation of the chain (think MMR root hash).
```

primitives/src/commitment.rs
```diff
@@ -0,0 +1,203 @@
...
+	/// For transition blocks it also MUST contain details of the next validator set.
+	pub payload: TPayload,
+
+	/// Finalized block number this commitment is for.
+	///
+	/// Grandpa validators agree on a block they create a commitment for and start collecting
```
andresilva, line 35:
```suggestion
	/// GRANDPA validators agree on a block they create a commitment for and start collecting
```

primitives/src/commitment.rs
```diff
@@ -0,0 +1,203 @@
...
+	/// signature. This process is called a round.
```
andresilva, line 36:
```suggestion
	/// signatures. This process is called a round.
```

primitives/src/commitment.rs
```diff
@@ -0,0 +1,203 @@
...
+	/// There might be multiple rounds in progress (depending on the block choice rule), however
+	/// since the payload is supposed to be cummulative, it is not required to import all
```
andresilva, line 38:
```suggestion
	/// since the payload is supposed to be cumulative, it is not required to import all
```

primitives/src/commitment.rs
```diff
@@ -0,0 +1,203 @@
...
+	/// commitments.
+	/// BEEFY light client is expected to import at least one commitment per epoch (the one with
+	/// [is_set_transition_block] set), but is free to import as many as it requires.
+	pub block_number: TBlockNumber,
+
+	/// BEEFY valitor set supposed to sign this comitment.
```
andresilva, line 44:
```suggestion
	/// BEEFY validator set supposed to sign this commitment.
```

primitives/src/commitment.rs
```diff
@@ -0,0 +1,203 @@
...
+	TPayload: cmp::Eq,
+{
+	fn cmp(&self, other: &Self) -> cmp::Ordering {
+		self.validator_set_id
+			.cmp(&other.validator_set_id)
+			.then_with(|| self.block_number.cmp(&other.block_number))
+	}
+}
+
+/// A commitment with matching Grandpa validators' signatures.
```
andresilva, line 82:
```suggestion
/// A commitment with matching GRANDPA validators' signatures.
```

primitives/src/commitment.rs
```diff
@@ -0,0 +1,203 @@
...
+#[derive(Debug, PartialEq, Eq, codec::Encode, codec::Decode)]
+pub struct SignedCommitment<TBlockNumber, TPayload, TSignature> {
+	/// The commitment signatures are collected for.
+	pub commitment: Commitment<TBlockNumber, TPayload>,
+	/// Grandpa validators' signatures for the commitment.
```
andresilva, line 87:
```suggestion
	/// GRANDPA validators' signatures for the commitment.
```

primitives/src/witness.rs
```diff
@@ -0,0 +1,131 @@
...
+/// This can be used by light clients for 2-phase interactive verification (for instance for
+/// Ethereum Mainnet), in a commit-reveal like scheme, where first we submit only the signed
+/// commitment witness and later on, the client picks only some signatures to verify at random.
+#[derive(Debug, PartialEq, Eq, codec::Encode, codec::Decode)]
+pub struct SignedCommitmentWitness<TBlockNumber, TPayload, TMerkleRoot> {
+	/// The full content of the commitment.
+	pub commitment: Commitment<TBlockNumber, TPayload>,
+
+	/// The bit vector of validators who signed the commitment.
+	pub signed_by: Vec<bool>, // TODO [ToDr] Consider replacing with bitvec crate
```
andresilva, line 42: 👍

primitives/tests/light_client/validator_set.rs
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) 2020 Parity Technologies (UK) Ltd.
```
andresilva, line 1: I guess this file will have some stuff related to the validator set in the future?

primitives/tests/light_client/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+		expected: ValidatorSetId,
+		got: ValidatorSetId,
+	},
+	/// [Commitment] can't be imported, cause it's a set transition block and the proof is missing.
+	InvalidValidatorSetProof,
+	/// [Commitment] is not useful, cause it's made for an older block than we know of.
+	///
+	/// In practice it's okay for the light client to import such commitments (if the validator set
+	/// matches), but it doesn't provide any more value, since the payload is meant to be
+	/// cummulative.
```
andresilva, line 62:
```suggestion
	/// cumulative.
```

primitives/tests/light_client/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+	/// It might be useful however, if we want to verify proofs that were generated against this
+	/// specific block number.
+	OldBlock {
+		/// Best block currently known by the light client.
+		best_known: BlockNumber,
+		/// Block in the commitment.
+		got: BlockNumber,
+	},
+	/// There is too many signatures in the commitment - more than validators.
```
andresilva, line 71:
```suggestion
	/// There are too many signatures in the commitment - more than validators.
```

primitives/tests/light_client/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+				got: no_of_non_empty_signatures,
+				valid: Some(valid),
+			});
+		}
+
+		Ok(commitment)
+	}
+
+	fn minimal_number_of_signatures(&self) -> usize {
+		(2 * self.validator_set.1.len() + 2) / 3
```
andresilva, line 222:
```suggestion
		(2 * self.validator_set.1.len() + 1) / 3
```

andresilva (approved):
LGTM, just commenting on typos. Currently we are coupling it quite a bit to GRANDPA, whereas BEEFY could work with other finality gadgets. This is probably a good idea for now as it makes it more accessible to people (we are going to use this with GRANDPA in the end), but in some cases we say stuff like "Commitment signed by the GRANDPA validators" where actually the commitment is signed by BEEFY validators (which happen to be the same ones as GRANDPA). I think the level of detail in the docs is probably appropriate for now (otherwise it might get too confusing).

primitives/tests/light_client/validator_set.rs
tomusdrw: Well, it contains the `Signature` and `Public` key of the validator set already, I don't expect there is going to be much more logic (we could encode set transitions here though if needed, but I find it simpler to put directly in the LC code for now).

primitives/tests/light_client/mod.rs
tomusdrw: Hmm, for `2` validators we still need both of them to sign, while `(2*2 + 1) / 3 = 1`, similarly for `5` we want at least 4, so `(2 * 5 + 2) / 3 = 4`

So I think this is correct.

andresilva: For N = 3 this method will return 2, whereas we need the 3 of them. I think the main issue (that I didn't notice earlier) is that we are adding the `+ 2` before doing the division. `2 * self.validator_set.1.len() / 3 + 1` should do the right thing.

primitives/tests/light_client/validator_set.rs
andresilva: I only asked this because of the file name 😛

primitives/tests/light_client/mod.rs
tomusdrw: You're right. I thought we need `ceil(2/3 * X)`, but we actually need `floor(2/3 * X + 1)`

andresilva (approved):

Issue #14, opened 2020-11-02
Change `is_set_transition_block` to an enum


tomusdrw:
Initially we're going to implement BEEFY key's changes per every session (see #9), so we need a differentiator between set change (bumps validator_set_id and contains new keys) and key change (the commitment contains new keys in some form).

We should change the field to be an enum, which is is also going to be more future proof.

tomusdrw:
Deemed not required in #21

Issue #9, opened 2020-10-22
GRANDPA keys change every session, but BEEFY keys shouldn't


tomusdrw:
Currently session is 4h on polkadot, and that's the cadence of GRANDPA keys changing.
The validator set however only changes every era (epoch), which is 4 sessions (around 16h).

We should make sure that BEEFY keys stay the same across the entire epoch (i.e. to have one set transition block per epoch, not per era).

CC @andresilva

seunlanlege:
dumb question, but wouldn't this mean that GRANDPA keys will diverge from BEEFY keys?

tomusdrw:
Session keys are already different for GRANDPA and BEEFY (it's even different crypto), but the validator set stays the same (i.e. the slashable actors). So this change is not about diverging on the validator set members level, but just the session keys they are using - i.e. for BEEFY they wouldn't change the session keys that frequently.

seunlanlege:
Right,wouldn't that bring up extra security concerns if the BEEFY session key never changes in between sessions?

Is this because there's a risk BEEFY might not complete a round in 4h?

tomusdrw:
> Right,wouldn't that bring up extra security concerns if the BEEFY session key never changes in between sessions?

This does not influence the security of the protocol (same bonded actors) - the point of changing session keys frequently is to reduce the risks in case they get compromised (i.e. validator leaks their keys and someone can use them to vote on alternative forks and/or trigger equivocations). So it makes it a bit less convenient (or more risky) for the validators.

> Is this because there's a risk BEEFY might not complete a round in 4h?

Not really, BEEFY should be completing rounds as quickly as possible assuming validators are online and network conditions are good, ideally every 2-3 blocks, similarly as GRANDPA.
The point is to reduce the minimal frequency of light client updates. Now it's once per session, because the session keys change - i.e. a light client has to import at least one BEEFY justification every session. With this reduction, a light client could only import stuff once per era which may significantly reduce the costs (obviously increasing latency though).

seunlanlege:
> So it makes it a bit less convenient (or more risky) for the validators.

Yeah this is what I meant.

> With this reduction, a light client could only import stuff once per era which may significantly reduce the costs (obviously increasing latency though).

Personally I think mandating that light clients import a new header every session (4h) is fine and 4h is already a high enough latency, 16h just makes the bridge unusable.

andresilva:
> Yeah this is what I meant.

[...]

PR #21, opened 2020-11-03
Support session change commitments

tomusdrw:
Ideally BEEFY keys should not change every session (see #9), but for now we will create a special kind of commitment which only indicates that the session keys are updated.

@andresilva I'm actually not sure if we really need this - from BEEFY perspective there isn't really any difference between changing the validator set and only changing public keys, I think we might safely use substrate's session index as `Commitment::validator_set_id`. The only drawback is that set ids won't match between GRANDPA and BEEFY, but does it really matter?

Closes: #14

### DIFF
--- a/primitives/src/commitment.rs
+++ b/primitives/src/commitment.rs
@@ -49,12 +49,11 @@ pub struct Commitment<TBlockNumber, TPayload> {
 	/// details are provided though.
 	pub validator_set_id: ValidatorSetId,
 
-	/// Indicator of the last block of the epoch.
+	/// Indicator of this commitment special purpose.
 	///
-	/// The payload will contain some form of the NEW validator set public keys information,
-	/// yet the block is signed by the current validator set.
-	/// When this commitment is imported, the client MUST increment the `validator_set_id`.
-	pub is_set_transition_block: bool,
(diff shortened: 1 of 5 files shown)


andresilva (commented):
I am also not entirely sure if this is needed. For example, GRANDPA doesn't make any distinction between a session or era change, if the keys change it starts a new set and increments the id (regardless of whether only session keys changes, or the actual staking authorities changed).

Here I think we'll have the same behavior, i.e. when we see some block that changes the authorities we know it must be committed (and we don't move forward before we do so), the only difference on session/era is whether we increment the set id or not which I don't think it's worth it to differentiate.

tomusdrw:
Cool! Thanks for confirmation, I'm closing this then along with #14

andresilva:
Right now the consensus protocols don't really know about eras or staking, they just interface with the session module and just check whether any keys have changed since the last session. Additionally there's also this `changed` boolean which the session module may set to true if the authority keys have not changed but the staking economics identities may have changed (e.g. more funds have been staked). Example: https://github.com/paritytech/substrate/blob/master/frame/grandpa/src/lib.rs#L605-L607

PR #22, opened 2020-11-09
Update to substrate master

andresilva:
Substrate 2.0 doesn't build with a recent rust nightly https://github.com/paritytech/substrate/pull/7238. We also want to track substrate master as we will require the MMR pallet.


andresilva:
Not sure what's going on with `cargo-deny`. I can replicate the problem locally using the latest version of `cargo-deny`.

PR #24, opened 2020-11-18
Integrate MMR and deposit root into the digest.

tomusdrw:
I'm considering making the MMR leafs a bit more alike polkadot to make it easy to use as an example, but that would be done in a separate PR.

### DIFF
--- a/node-example/runtime/Cargo.toml
+++ b/node-example/runtime/Cargo.toml
@@ -28,51 +28,52 @@ serde = { features = ['derive'], optional = true, version = '1.0.117' }
 pallet-template = { path = '../pallets/template', default-features = false, version = '2.0.0' }
 
 # Substrate dependencies
-frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false, optional = true }
-frame-executive = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-frame-support = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-frame-system = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-frame-system-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false, optional = true }
-frame-system-rpc-runtime-api = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-pallet-aura = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-pallet-balances = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-pallet-grandpa = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-pallet-randomness-collective-flip = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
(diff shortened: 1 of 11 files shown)


node-example/runtime/Cargo.toml
```diff
@@ -30,31 +30,32 @@ pallet-beefy = { path = "../../pallet", default-features = false }
...
-sp-consensus-aura = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-sp-core = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-sp-inherents = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-sp-offchain = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-sp-session = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-sp-std = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-sp-transaction-pool = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
-sp-version = { git = "https://github.com/paritytech/substrate", branch = "master", default-features = false }
+frame-benchmarking = { git = "https://github.com/paritytech/substrate", default-features = false, optional = true , branch = "td-mmr" }
```
andresilva, line 33: There is an extra space in all dependencies with `default-features = false , branch = "td-mmr"`. I guess this is temporary anyway, since we will switch to master once MMR is merged on substrate.

node-example/runtime/src/lib.rs
```diff
@@ -274,6 +274,36 @@ impl pallet_template::Trait for Runtime {
...
+pub struct DepositLog;
+impl pallet_mmr::primitives::OnNewRoot<MmrHash> for DepositLog {
+	fn on_new_root(root: &Hash) -> Weight {
+		let digest = DigestItem::Consensus(
+			beefy_primitives::BEEFY_ENGINE_ID,
+			codec::Encode::encode(&beefy_primitives::ConsensusLog::<BeefyId>::MmrRoot(*root)),
+		);
+		<frame_system::Module<Runtime>>::deposit_log(digest);
+
+		<Runtime as frame_system::Trait>::DbWeight::get().writes(1)
```
andresilva, line 293: This was a bit surprising to me but indeed `fn deposit_log` is documented as being `O(1)` and performing 1 storage write. Not sure if this is accurate though since the storage write is only ephemeral (the entry is killed on block finalization) and the changes aren't actually committed to disk. Anyway I think this is fine as is :)

andresilva (approved):

node-example/runtime/Cargo.toml
tomusdrw: Updated.

node-example/runtime/src/lib.rs
tomusdrw: Yeah, this thing is flawed anyway, I'm currently rewriting substrate PR to actually have `on_initialize` weight benchmarked instead of hardcoded, so will update here soon as well.

andresilva (approved):

PR #25, opened 2020-11-18
Add beefy pallet

andresilva:
Add the beefy pallet that maintains the authority set and integrates with the session module.


beefy-gadget/src/lib.rs
```diff
@@ -336,43 +325,22 @@ pub async fn start_beefy_gadget<Block, Backend, Client, Network, SyncOracle>(
...
-
-	let voters = voters
-		.into_iter()
-		.map(|address| AuthorityId::from_string(address).unwrap())
-		.collect::<Vec<_>>();
+	let at = BlockId::hash(client.info().best_hash);
+	let voters = client
+		.runtime_api()
+		.authorities(&at)
+		.expect("Failed to get BEEFY authorities");
```
tomusdrw, line 332: Will it ever happen?

node-example/runtime/Cargo.toml
```diff
@@ -55,42 +57,44 @@ sp-transaction-pool = { git = "https://github.com/paritytech/substrate", branch
...
-    'sp-core/std',
-    'sp-inherents/std',
-    'sp-offchain/std',
-    'sp-runtime/std',
-    'sp-session/std',
-    'sp-std/std',
-    'sp-transaction-pool/std',
-    'sp-version/std',
+	"codec/std",
+	"serde",
```
tomusdrw, line 73: Could you sort both of these lists?

pallet/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "beefy-pallet"
```
tomusdrw, line 2:
```suggestion
name = "pallet-beefy"
```
I think that's more in line with other pallets.

pallet/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl<T: Trait> pallet_session::OneSessionHandler<T::AccountId> for Module<T> {
+	type Key = T::AuthorityId;
+
+	fn on_genesis_session<'a, I: 'a>(validators: I)
+	where
+		I: Iterator<Item = (&'a T::AccountId, T::AuthorityId)>,
+	{
+		let authorities = validators.map(|(_, k)| k).collect::<Vec<_>>();
+		Self::initialize_authorities(&authorities);
```
tomusdrw, line 83: Wouldn't that panic in case we have authorities in the chain spec already?

primitives/src/lib.rs
```diff
@@ -32,7 +35,57 @@
...
+
+	/// Identity of a BEEFY authority using ECDSA as its crypto.
+	pub type AuthorityId = app_ecdsa::Public;
+
+	/// Signature for a BEEFY authority using ECDSA as its crypto.
+	pub type AuthoritySignature = app_ecdsa::Signature;
+}
+
+/// The `ConsensusEngineId` of BEEFY.
+pub const BEEFY_ENGINE_ID: sp_runtime::ConsensusEngineId = *b"BEEF";
```
tomusdrw, line 66: is upper case notation used for other consensus engines?

tomusdrw (approved):
lgtm!

beefy-gadget/src/lib.rs
andresilva: I think this will only be used to fetch genesis authorities. Once we add support for changing the BEEFY authorities then we will be tracking the current authority set externally based on the header digests.

primitives/src/lib.rs
andresilva: Looking at `ConsensusEngineId` in substrate it seems that there's no standard 🙈 Half are uppercase and the other half is lowercase. I don't have a strict opinion on this, can rename it to `beef` :)

pallet/Cargo.toml
andresilva: 👍

pallet/src/lib.rs
andresilva: We guard against this inside `initialize_authorities` and only set the authorities if they're not set already. In the case where we have authorities defined in the module genesis config and also on genesis session I don't know which would take precedence, I think it would depend on module order when assembling runtime.

primitives/src/lib.rs
tomusdrw: Yeah, I'm fine with either, was just asking if there is a convention here.

pallet/src/lib.rs
andresilva: I took this code from the aura pallet and misread it now when you commented, so my comment above doesn't make much sense. The code will indeed panic if there are authorities defined both on the module genesis config and as part of genesis session. This is to make sure the end-user doesn't try to initialize a chain using both mechanisms.

PR #26, opened 2020-11-18
Signed commitments rpc pubsub

andresilva:
Merge after #25.

Adds an RPC pubsub method for streaming signed commitments.


beefy-gadget/rpc/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+		let stream = self
+			.signed_commitment_stream
+			.subscribe()
+			.map(|x| Ok::<_, ()>(notification::SignedCommitment::new::<Block, Signature>(x)))
+			.map_err(|e| warn!("Notification stream error: {:?}", e))
+			.compat();
+
+		self.manager.add(subscriber, |sink| {
+			#[allow(clippy::redundant_closure)]
+			let stream = stream.map(|res| Ok(res));
```
tomusdrw, line 106:
```suggestion
			let stream = stream.map(Ok);
```
?

beefy-gadget/rpc/src/notification.rs
```diff
@@ -0,0 +1,36 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use codec::Encode;
+use serde::{Deserialize, Serialize};
+
+use sp_runtime::traits::Block as BlockT;
+
+/// An encoded signed commitment proving that the given header has been finalized.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct SignedCommitment(sp_core::Bytes);
```
tomusdrw, line 24: Can we add a documentation pointer to the type that is expected to be encoded here? I presume it's `primitives::SignedCommitment`?

beefy-gadget/src/lib.rs
```diff
@@ -231,14 +231,21 @@ where
 	}
 
 	fn handle_finality_notification(&mut self, notification: FinalityNotification<Block>) {
-		info!(target: "beefy", "Finality notification: {:?}", notification);
+		debug!(target: "beefy", "🥩 Finality notification: {:?}", notification);
 
 		if self.should_vote_on(*notification.header.number()) {
+			let commitment = Commitment {
+				payload: notification.header.hash(),
```
tomusdrw, line 238: This should be configurable, we want this to be a MMR root. I'd put a TODO in all places like that, but it's fine :)

beefy-gadget/src/lib.rs
```diff
@@ -231,14 +231,21 @@ where
...
+				block_number: notification.header.number(),
+				validator_set_id: 0,
```
tomusdrw, line 240: Same here, this should be filled in in the future :)

tomusdrw (approved):
:tada:

tomusdrw (approved):

PR #27, opened 2020-12-01
use new mmr root as commitment payload

andresilva:
Currently just reading from `ConsensusLog` digest.

### DIFF
--- a/beefy-gadget/src/lib.rs
+++ b/beefy-gadget/src/lib.rs
@@ -24,7 +24,7 @@ use futures::{future, FutureExt, Stream, StreamExt};
 use log::{debug, error, info, trace, warn};
 use parking_lot::Mutex;
 
-use beefy_primitives::{BeefyApi, Commitment, ConsensusLog, SignedCommitment, BEEFY_ENGINE_ID, KEY_TYPE};
+use beefy_primitives::{BeefyApi, Commitment, ConsensusLog, MmrRootHash, SignedCommitment, BEEFY_ENGINE_ID, KEY_TYPE};
 
 use sc_client_api::{Backend as BackendT, BlockchainEvents, FinalityNotification, Finalizer};
 use sc_network_gossip::{
@@ -35,7 +35,6 @@ use sp_api::{BlockId, ProvideRuntimeApi};
 use sp_application_crypto::{AppPublic, Public};
 use sp_blockchain::HeaderBackend;
 use sp_consensus::SyncOracle as SyncOracleT;
-use sp_core::H256;
 use sp_keystore::{SyncCryptoStore, SyncCryptoStorePtr};
 use sp_runtime::{
 	generic::OpaqueDigestItemId,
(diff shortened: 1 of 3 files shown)


beefy-gadget/src/lib.rs
```diff
@@ -165,7 +169,7 @@ struct BeefyWorker<Block: BlockT, Id, Signature, FinalityNotifications> {
 	local_id: Id,
 	key_store: SyncCryptoStorePtr,
 	min_interval: u32,
-	rounds: Rounds<Block::Hash, NumberFor<Block>, Id, Signature>,
+	rounds: Rounds<H256, NumberFor<Block>, Id, Signature>,
```
tomusdrw, line 172: Could we do `type MmrRootHash = H256;`? It took me a while to figure out what that `H256` here is.

beefy-gadget/src/lib.rs
```diff
@@ -231,13 +235,20 @@ where
...
-		debug!(target: "beefy", "🥩 Finality notification: {:?}", notification);
+		debug!(target: "beefy", "Finality notification: {:?}", notification);
 
 		if self.should_vote_on(*notification.header.number()) {
-			// TODO: this needs to be properly populated by signing an MMR root as the payload
-			// (and/or abstracting the "thing to sign") and with support for validator set changes.
+			let mmr_root = if let Some(hash) = find_mmr_root_digest::<Block, Id>(&notification.header) {
+				hash
+			} else {
+				debug!(target: "beefy", "No MMR root digest found for: {:?}", notification.header.hash());
```
tomusdrw, line 244: Maybe make it a warning?
```suggestion
				warn!(target: "beefy", "🥩 No MMR root digest found for: {:?}", notification.header.hash());
```

beefy-gadget/src/lib.rs
```diff
@@ -231,13 +235,20 @@ where
 	}
 
 	fn handle_finality_notification(&mut self, notification: FinalityNotification<Block>) {
-		debug!(target: "beefy", "🥩 Finality notification: {:?}", notification);
+		debug!(target: "beefy", "Finality notification: {:?}", notification);
```
tomusdrw, line 238: I liked the icon, any reason for removing? :)

andresilva: I kept it only for "user-level" messages, i.e. info/warn.

primitives/src/lib.rs
```diff
@@ -82,8 +85,8 @@ pub enum ConsensusLog<AuthorityId: Codec> {
 	#[codec(index = "2")]
 	OnDisabled(AuthorityIndex),
 	/// MMR root hash.
-	#[codec(index = "2")]
-	MmrRoot(H256),
+	#[codec(index = "3")]
```
tomusdrw, line 88: Good catch :+1:

tomusdrw (approved):

PR #32, opened 2020-12-09
Update to latest substrate.

tomusdrw:


node-example/runtime/src/lib.rs
```diff
@@ -127,21 +126,36 @@ pub fn native_version() -> NativeVersion {
...
-	/// We allow for 2 seconds of compute with a 6 second average block time.
-	pub const MaximumBlockWeight: Weight = 2 * WEIGHT_PER_SECOND;
-	pub const AvailableBlockRatio: Perbill = Perbill::from_percent(75);
-	/// Assume 10% of weight for average on_initialize calls.
-	pub MaximumExtrinsicWeight: Weight = AvailableBlockRatio::get()
-		.saturating_sub(Perbill::from_percent(10)) * MaximumBlockWeight::get();
-	pub const MaximumBlockLength: u32 = 5 * 1024 * 1024;
+	pub BlockLength: limits::BlockLength = limits::BlockLength
+		::max_with_normal_ratio(5 * 1024 * 1024, NORMAL_DISPATCH_RATIO);
+	pub BlockWeights: limits::BlockWeights = limits::BlockWeights::builder()
```
andresilva, line 137: I am not familiar with these changes, looking at substrate node we have an extra call here to `.base_block(BlockExecutionWeight::get())`, not sure if we just want to replicate what we have there.

andresilva (approved):

node-example/runtime/src/lib.rs
tomusdrw: It's a default value anyway.

PR #47, opened 2021-01-14
Intro and walkthrough documentation

andresilva:
This is the document that I prepared for our call. I just converted it to markdown from org-mode. If you feel like this shouldn't exist on the repo I can just add it to Notion instead.


adoerr (approved):
The repo is exactly the right place! Thanks again for the intro and the write up!

tomusdrw:
we should also consider updating the readme, cause it's slightly out of date :)

PR #71, opened 2021-01-25
Bump Substrate and Deps

adoerr:
This bump will also include the improved GRANDPA finality RPC API [Commit](https://github.com/paritytech/substrate/commit/bd5c9a646896b11061a6ff92af3c673958e238eb)


adoerr:
cargo deny complains about `raw-cpuid` version, which should be updated to `9.0.0`. This is a dependency of `cranelift-native`. A pin to version `9.0.0` fails.

So, I would overrule cargo deny for this one.

beefy-gadget/Cargo.lock
```diff
@@ -1,4321 +0,0 @@
-# This file is automatically @generated by Cargo.
```
tomusdrw, line 1: Is it already part of `.gitignore`? If not, maybe let's add it?

tomusdrw (approved):
@adoerr yes, please add an explicit ignore (with a comment) in `deny.toml` file, here: https://github.com/paritytech/grandpa-bridge-gadget/blob/acbce750a87a2ad4a27f0a4900ff93659d1a44b0/deny.toml#L52

PR #73, opened 2021-01-28
Bump Substrate

adoerr:
Also removed old Sec advisory.


Cargo.lock
```diff
@@ -5911,18 +5911,18 @@ checksum = "388a1df253eca08550bef6c72392cfe7c30914bf41df5269b68cbd6ff8f570a3"
 
 [[package]]
 name = "serde"
-version = "1.0.122"
+version = "1.0.119"
```
HCastano: Why did `serde` need to be downgraded?

Cargo.lock
```diff
@@ -6900,9 +6900,9 @@ checksum = "1e81da0851ada1f3e9d4312c704aa4f8806f0f9d69faaf8df2f3464b4a9437c2"
 
 [[package]]
 name = "syn"
-version = "1.0.60"
+version = "1.0.58"
```
HCastano: Same question here

deny.toml
```diff
@@ -48,8 +48,6 @@ notice = "warn"
 # A list of advisory IDs to ignore. Note that ignored advisories will still
 # output a note when they are encountered.
 ignore = [
-	# yaml-rust < clap. Not feasible to upgrade and also not possible to trigger in practice.
-	"RUSTSEC-2018-0006",
```
HCastano, line 52: If it's fine to get rid of this, we should probably also do it in the `parity-bridges-common` repo

HCastano (approved):

Cargo.lock
adoerr: I have no idea

Cargo.lock
adoerr: Still no idea

PR #74, opened 2021-02-01
Beefy pallet test

adoerr:
This PR provides a basic pallet test infrastructure before migrating the Beefy pallet to FRAMEv2.

### DIFF
--- a/beefy-pallet/Cargo.toml
+++ b/beefy-pallet/Cargo.toml
@@ -13,9 +13,15 @@ beefy-primitives = { path = "../primitives", default-features = false }
 
 frame-support = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 frame-system = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
+
 sp-runtime = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 sp-std = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
+
 pallet-session = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
+
+[dev-dependencies]
+sp-core = { git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-io = { git = "https://github.com/paritytech/substrate.git", branch = "master" }
 
 [features]
 default = ["std"]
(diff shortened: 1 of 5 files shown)


beefy-pallet/src/tests.rs
```diff
@@ -0,0 +1,83 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+#![cfg(test)]
```
tomusdrw, line 17:
```suggestion
```
likewise

beefy-pallet/src/tests.rs
```diff
@@ -0,0 +1,83 @@
...
+		assert_eq!(want[1], authorities[1]);
+	});
+}
+
+#[test]
+fn session_change_updates_authorities() {
+	new_test_ext(vec![1, 2, 3, 4]).execute_with(|| {
+		init_block(1);
+
+		assert!(0 == Beefy::validator_set_id());
```
tomusdrw, line 65: Might be worth checking `Beefy::authorities()` and `Beefy::next_authorities()` here. I think the next storage entry should be updated after genesis.

tomusdrw (commented):
lgtm! couple of minor things

beefy-pallet/Cargo.toml
```diff
@@ -13,10 +13,17 @@ beefy-primitives = { path = "../primitives", default-features = false }
...
 frame-support = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 frame-system = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
+
 sp-runtime = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 sp-std = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
+
 pallet-session = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 
+[dev-dependencies]
+sp-core = { git = "https://github.com/paritytech/substrate.git", branch = "master" }
```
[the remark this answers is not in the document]
adoerr: I don;t think so, as long as `sp-core` is not used in parallel as a dependency in the same crate. Besides, it seems to be common practice. I checked with a couple of other pallets.

beefy-pallet/src/tests.rs
adoerr: I have added a separate test for `Beefy::next_authorities()` update. The way `OneSessionHandler` is currently implemented, next authorities *only* update if authorities themselves update as well. In other words, for session `0` and `1` both, authorities and next authorities don't update. Let's discuss this during the upcoming sync-up.

tomusdrw (approved):
lgtm!

PR #76, opened 2021-02-03
Migrate beefy-pallet to FRAMEv2

adoerr:


beefy-pallet/src/lib.rs
```diff
@@ -35,37 +35,73 @@ mod mock;
...
+		}
+	}
 }
 
-impl<T: Config> Module<T> {
+impl<T: Config> Pallet<T> {
 	fn change_authorities(new: Vec<T::AuthorityId>, queued: Vec<T::AuthorityId>) {
-		// As in GRANDPA, we don't trigger validator set change if the set actually
-		// remains the same.
+		// As in GRANDPA, we trigger a validator set change only, the the validator
```
HCastano, line 100:
```suggestion
		// As in GRANDPA, we trigger a validator set change only if the the validator
```

HCastano (approved):
Now might be a good time to add some module level comments for the pallet, but it can be done in a follow up if you want.

Also, did you check the metadata to make sure everything stayed the same? I imagine it's not a big deal since this isn't deployed anywhere, but I'm just curious.

adoerr:
Good point about the module level comments. Would like to postpone that a bit still. I do expect quite a bunch of changes coming up.

I did not bother with the Metadata. How would I check that anyways? Using polka.js and then 'Settings' -> 'Metadata'?

Issue #40, opened 2021-01-07
Make sure the services works with no-validators


tomusdrw:
It should simply receive gossip messages, but don't do any voting.

adoerr:
A straightforward fix for this should be to make `BeefyWorker.local_id` optional and call `BeefyWorker.should_vote_on()` only if `local_id.is_some()`.

PR #77, opened 2021-02-05
Run BEEFY worker as non-validator

adoerr:
Reselves: #40

### DIFF
--- a/beefy-gadget/src/lib.rs
+++ b/beefy-gadget/src/lib.rs
@@ -164,13 +164,13 @@ struct VoteMessage<Hash, Number, Id, Signature> {
 	signature: Signature,
 }
 
-enum BeefyValidator<Id> {
+enum BeefyId<Id> {
 	Validator(Id),
-	NoValidator,
+	None,
 }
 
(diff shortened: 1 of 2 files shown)


beefy-gadget/src/lib.rs
```diff
@@ -164,8 +164,13 @@ struct VoteMessage<Hash, Number, Id, Signature> {
 	signature: Signature,
 }
 
+enum BeefyValidator<Id> {
```
tomusdrw, line 167:
```suggestion
enum BeefyWorkerMode<Id> {
```

I'd change the naming, cause currently `validator` is duplicated in both variants (and the type itself) and `BeefyValidator::NoValidator` looks a bit weird :)

tomusdrw (approved):
looks good!

beefy-gadget/src/lib.rs
adoerr: How about changing the type name to `BeefyAuthority`? This would give us `BeefyAuthority::NoValidator` and `BeefyAuthority::Validator`.

PR #80, opened 2021-02-08
Add BEEFY gadget as extra peer set

adoerr:


tomusdrw (approved):
LGTM, thanks!

it would be really good to somehow have a end-to-end test for this, basically running two nodes and making sure they produce justifications (subscription). I'll open a separate issue for that.

PR #85, opened 2021-02-11
Add ValidatorSetId to BEEFY digest

adoerr:
Adding `ValidatorSetId` to the BEEFY digest will allow for easier validator set changes tracking within the BEEFY gadget.

### DIFF
--- a/beefy-primitives/src/lib.rs
+++ b/beefy-primitives/src/lib.rs
@@ -80,7 +80,12 @@ pub type MmrRootHash = H256;
 pub enum ConsensusLog<AuthorityId: Codec> {
 	/// The authorities have changed.
 	#[codec(index = 1)]
-	AuthoritiesChange((Vec<AuthorityId>, ValidatorSetId)),
+	AuthoritiesChange {
+		/// Set of new validators to be used
+		new_validator_set: Vec<AuthorityId>,
+		/// Id for this new set of validators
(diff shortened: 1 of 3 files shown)


beefy-primitives/src/lib.rs
```diff
@@ -80,7 +80,7 @@ pub type MmrRootHash = H256;
 pub enum ConsensusLog<AuthorityId: Codec> {
 	/// The authorities have changed.
 	#[codec(index = 1)]
-	AuthoritiesChange(Vec<AuthorityId>),
+	AuthoritiesChange((Vec<AuthorityId>, ValidatorSetId)),
```
tomusdrw, line 83:
```suggestion
	AuthoritiesChange(Vec<AuthorityId>, ValidatorSetId),
```

Instead of having one-value enum with a tuple instead, we can just do two-valued enum

tomusdrw: You can also consider naming the parameters to add some extra docs:
```rust
AuthoritiesChange {
   /// ...
   new_validator_set: Vec<AuthorityId>,
   /// ...
   new_validator_set_id: ValidatorSetId,
}
```

tomusdrw (commented):
Looks good except for the redundant tuple!

tomusdrw (approved):

PR #89, opened 2021-02-23
Offchain MMR storage key.

tomusdrw:


adoerr (approved):
LGTM.

What is the purpose of this storage key command? Being able to manually inject a node into the MMR offchain storage?

tomusdrw:
> What is the purpose of this storage key command? Being able to manually inject a node into the MMR offchain storage?

Rather to read individual leaves over RPC.

PR #91, opened 2021-02-26
Bump Substrate and Deps

adoerr:
Bump Substrate before work on BEEFY resumes. Also there have been some GRANDPA implementation changes recently.

### DIFF
--- a/beefy-gadget/src/lib.rs
+++ b/beefy-gadget/src/lib.rs
@@ -284,21 +284,22 @@ where
 				validator_set_id: 0,
 			};
 
+			// TODO: this is really ugly due to the recent signature change of `sign_with()`.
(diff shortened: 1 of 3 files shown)


tomusdrw (approved):

beefy-gadget/src/lib.rs
```diff
@@ -284,15 +284,20 @@ where
 				validator_set_id: 0,
 			};
 
+			// TODO: this is really ugly due to the recent signature change of `sign_with()`.
```
tomusdrw, line 287: Can you add an issue number to this `TODO`? So that it's not forgotten?

adoerr: Done #92

PR #94, opened 2021-03-02
Track BEEFY validator set

### LINKED ISSUE #15: Track BEEFY validator set
Get the client code to a point where it signs some arbitrary data.
This most likely includes having the gadget to track BEEFY validator set somehow.

adoerr:
Resolves #15

### DIFF
--- a/beefy-gadget/src/lib.rs
+++ b/beefy-gadget/src/lib.rs
@@ -25,7 +25,8 @@ use log::{debug, error, info, trace, warn};
 use parking_lot::Mutex;
 
 use beefy_primitives::{
-	BeefyApi, Commitment, ConsensusLog, MmrRootHash, SignedCommitment, ValidatorSet, BEEFY_ENGINE_ID, KEY_TYPE,
+	BeefyApi, Commitment, ConsensusLog, MmrRootHash, SignedCommitment, ValidatorSet, ValidatorSetId, BEEFY_ENGINE_ID,
+	KEY_TYPE,
 };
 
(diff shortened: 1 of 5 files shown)


tomusdrw (commented):
lgtm :)

adoerr:
@tomusdrw do you prefer to have the draft being merged right away and work on validator set persistence in a separate PR?

tomusdrw:
> I could use just some static var for the time being in order to keep track of the current validator set id.

Yes, I think that's the best way. If the static var is empty (i.e. we didn't observe the digest yet), let's early exit and skip that session for now.

beefy-gadget/src/lib.rs
```diff
@@ -194,6 +197,7 @@ struct BeefyWorker<Block: BlockT, Id, Signature, FinalityNotifications> {
 	signed_commitment_sender: BeefySignedCommitmentSender<Block, Signature>,
 	best_finalized_block: NumberFor<Block>,
 	best_block_voted_on: NumberFor<Block>,
+	validator_set_id: ValidatorSetId,
```
tomusdrw, line 200: I think this should be an `Option` and we should skip voting if it's not set. Otherwise I'm afraid that BEEFY validators will start signing bullshit commitments. Later on that should obviously be penalized, but currently I think it's safer to just wait for the beginning of the session. Not sure how this will work initially though (i.e. if we ever produce the first commitment).

tomusdrw (approved):
lgtm, but I wonder if it's not safer to skip voting if we don't really know current validator set id.

PR #96, opened 2021-03-03
Ignore generic-array advisory.

tomusdrw:


adoerr (approved):
LGTM. I actually fixed that with #94 already.

tomusdrw:
Ah, cool :) I was surprised why `master` is failing and your PR is not.

PR #97, opened 2021-03-03
Ad 92

### LINKED ISSUE #92: Use CryptoStore instead of SyncCryptoStore
`beefy-gadget` is still using [SyncCryptoStore](https://github.com/paritytech/grandpa-bridge-gadget/blob/7ee10364794886848a39b3bd03fab168ec0aeeb4/beefy-gadget/src/lib.rs#L287-L301) which is deprecated.

`sp_keystore::CryptoStore` should be used instead. Also the signature of the key store methods has been changed recently [Make keystore return `None` when a key doesn't exist](https://github.com/paritytech/substrate/commit/59494855205374e8f27fb131425dfd78897a9298l). This change leads to a quite convoluted result handling in the current code which should be improved.

adoerr:
Resolves #92 

Using `CryptoStore` doesn't make sense currently. The only implementation of the `CryptoStore` trait is itself using `SyncCryptoStore` in it's implementation.

So I just streamlined the result handling.

### DIFF
--- a/beefy-gadget/src/lib.rs
+++ b/beefy-gadget/src/lib.rs
@@ -258,6 +258,16 @@ where
 		number == next_block_to_vote_on
 	}
 
+	fn sign_commitment(&self, id: &Id, commitment: &[u8]) -> Option<Signature> {
+		let sig = SyncCryptoStore::sign_with(&*self.key_store, KEY_TYPE, &id.to_public_crypto_pair(), &commitment)
+			.ok()
+			.flatten()?
+			.try_into()
+			.ok()?;
+
+		Some(sig)
+	}
+
 	fn handle_finality_notification(&mut self, notification: FinalityNotification<Block>) {
 		debug!(target: "beefy", "Finality notification: {:?}", notification);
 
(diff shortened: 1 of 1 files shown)


beefy-gadget/src/lib.rs
```diff
@@ -264,6 +264,16 @@ where
 		number == next_block_to_vote_on
 	}
 
+	fn sign_commitment(&self, id: &Id, commitment: &[u8]) -> Option<Signature> {
+		let sig = SyncCryptoStore::sign_with(&*self.key_store, KEY_TYPE, &id.to_public_crypto_pair(), &commitment)
+			.ok()
+			.flatten()?
+			.try_into()
+			.ok()?;
```
tomusdrw, line 272: Wouldn't it make sense to get a bit more details on why it failed (in case it fails in the future)? :)

Like maybe return a `Result<Signature, String>` from `sign_commitment` just to print it out in the warn in L316.

tomusdrw (approved):
lgtm, although to future-proof the code, I'd print more detailed error.

beefy-gadget/src/lib.rs
adoerr: Yes, it does make sense and it is exactly why the `SyncCryptoStore` API is totally broken, IMO. If we want the error root cause, we better leave the original code in place.

Otherwise, we have to introduce our own BEEFY error type and `map_err` the various __different__  error type that could occur onto our BEEFY error. Having a BEEFY error type is something, we most likely will end up anyway. But that is something I would like to do in a separate PR.

So add another issue for a BEEFY error type or dump the PR and leave the initial code?

PR #99, opened 2021-03-04
Add error type

### LINKED ISSUE #98: Add BEEFY error type
Introduce BEEFY error type for unified error handling.

The BEEFY error type should allow to convert dependency specific errors into BEEFY errors.

adoerr:
Resolves #98

Add proper error handling to `BeefyWorker::sign_commitment()`.

### DIFF
--- a/beefy-gadget/src/error.rs
+++ b/beefy-gadget/src/error.rs
@@ -1,0 +1,9 @@
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+	/// Check signature error
+	#[error("Message signature {0:?} by {1:?} is invalid.")]
+	InvalidSignature(Vec<u8>, Vec<u8>),
+	/// Sign commitment error
+	#[error("Failed to sign comitment using key: {0:?}. Reason: {1}")]
+	CannotSign(Vec<u8>, String),
+}
--- a/beefy-gadget/src/error.rs
+++ b/beefy-gadget/src/error.rs
@@ -1,9 +1,32 @@
+// Copyright (C) 2020-2021 Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: GPL-3.0-or-later WITH Classpath-exception-2.0
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use std::fmt::Debug;
+
+use sp_core::crypto::Public;
+
+/// BEEFY gadget specific errors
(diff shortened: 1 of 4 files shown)


beefy-gadget/src/error.rs
```diff
@@ -0,0 +1,9 @@
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
tomusdrw, line 2: Missing license pre-amble and whole enum docs.

beefy-gadget/src/error.rs
```diff
@@ -0,0 +1,9 @@
...
+	/// Check signature error
+	#[error("Message signature {0:?} by {1:?} is invalid.")]
```
tomusdrw, line 4: How long is the `Vec<u8>` here? can we maybe hex-convert before displaying? Otherwise it will be really hard to read.

beefy-gadget/src/error.rs
```diff
@@ -0,0 +1,9 @@
...
+	InvalidSignature(Vec<u8>, Vec<u8>),
+	/// Sign commitment error
+	#[error("Failed to sign comitment using key: {0:?}. Reason: {1}")]
+	CannotSign(Vec<u8>, String),
```
tomusdrw, line 8: same here. Debug impl of `Vec<u8>` is quite verbose :)

beefy-gadget/src/lib.rs
```diff
@@ -264,15 +268,17 @@ where
 		number == next_block_to_vote_on
 	}
 
-	fn sign_commitment(&self, id: &Id, commitment: &[u8]) -> Option<Signature> {
+	fn sign_commitment(&self, id: &Id, commitment: &[u8]) -> Result<Signature, Error> {
 		let sig = SyncCryptoStore::sign_with(&*self.key_store, KEY_TYPE, &id.to_public_crypto_pair(), &commitment)
-			.ok()
-			.flatten()?
+			.map_err(|e| Error::CannotSign(id.to_raw_vec(), e.to_string()))?
```
tomusdrw, line 273: I suspect you don't want to make the `Error` generic, hence vec conversion right?

beefy-gadget/src/lib.rs
```diff
@@ -44,10 +44,14 @@ use sp_runtime::{
 	traits::{Block as BlockT, Hash as HashT, Header as HeaderT, NumberFor, Zero},
 };
 
+mod error;
+
 pub mod notification;
 
 use notification::BeefySignedCommitmentSender;
 
+pub use error::Error;
```
tomusdrw, line 53: Shall we also export `Result`?  I think this is a common practice to use `my_lib::Result<()>` instead of `Result<(), my_lib::Error>`.
Note that I'm very much against importing `my_lib::Result` to the scope and overriding `std::result::Result` though (i.e. only full path usage).

adoerr: I made the error type `pub(crate)` for the time being, since currently it is really only used for `beefy_gadget` internal error handling.

beefy-gadget/src/error.rs
adoerr: Totally forgot about the header

beefy-gadget/src/error.rs
adoerr: Signature is now hex-converted to a string

beefy-gadget/src/lib.rs
adoerr: Error type is now generic of `Id`.

tomusdrw (approved):

Issue #95, opened 2021-03-02
Determine validator_set_id at BeefyWorker startup


adoerr:
`BeefyWorker` needs to be initialized with the current valid `validator_set_id`. Right now this id is hard coded to the genesis validator set id '0' (zero).

adoerr:
> I think this should be an Option and we should skip voting if it's not set. Otherwise I'm afraid that BEEFY validators will start signing bullshit commitments. Later on that should obviously be penalized, but currently I think it's safer to just wait for the beginning of the session. Not sure how this will work initially though (i.e. if we ever produce the first commitment) - perhaps special casing empty chain will be required (setting Some(0) instead of None).

Comment from @tomusdrw on #94.

Currently we always use `Some(0)` as the validator_set_id, otherwise the test deployment would never vote on anything. Reason is that the test deployment always starts with genesis, so we need to implement the empty chain test at the very least for the test deployments.

adoerr:
Determining the current active BEEFY `validator_set_id` would involve the following steps:

1. fetch the last finalized block number
`HeaderBackend::info()::finalized_number`
2. get the header for that block number
`HeaderBackend::header(number)`
3. scan the header for a BEEFY `AuthoritiesChange` digest
4. if found -> use validator set
5. if not found -> decrement header number; goto 2

special case handling is needed for genesis -> probably separate PR

PR #111, opened 2021-03-09
Initialize BeefyWorker with current validator set

adoerr:
Resolves: #95 

Currently we use BEEFY validator set for `BeefyWorker` initialization only, i.e. the validator set is not retained by the `BeefyWorker`. This is likely going to change in the future.

### DIFF
--- a/beefy-gadget/src/lib.rs
+++ b/beefy-gadget/src/lib.rs
@@ -16,9 +16,9 @@
 
 use std::{convert::TryFrom, fmt::Debug, sync::Arc};
 
-use {codec::Codec, log::info};
+use codec::Codec;
 
-use beefy_primitives::{BeefyApi, KEY_TYPE};
+use beefy_primitives::BeefyApi;
 
 use {
 	sc_client_api::{Backend as BackendT, BlockchainEvents, Finalizer},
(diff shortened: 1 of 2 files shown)


beefy-gadget/src/lib.rs
```diff
@@ -111,40 +111,23 @@ pub async fn start_beefy_gadget<Block, Pair, Backend, Client, Network, SyncOracl
...
-		Some(id) => {
-			info!(target: "beefy", "🥩 Starting BEEFY worker with local id: {:?}", id);
-			worker::BeefyId::Validator(id.clone())
-		}
-		None => {
-			info!(target: "beefy", "🥩 No local id found, BEEFY worker will be gossip only.");
-			worker::BeefyId::None
-		}
-	};
+		.expect("Failed to get BEEFY validator set");
```
tomusdrw, line 118: So this obviously works for now, but what if the client is not fully in sync with the network? I think we need a bit more sophisticated mechanism, where we actually update the validator set id when new blocks are imported / we get synced up.

tomusdrw: Also regarding `expect` - the assumption here is that the beefy pallet is part of the runtime (pretty much since genesis), however existing changes will most likely upgrade to beefy pallet, so I think we should be handling this error a bit more gracefuly.

Could you log issues to make sure this is covered?

tomusdrw (approved):
lgtm modulo the sync/genesis grumbles.

beefy-gadget/src/lib.rs
adoerr: For the out-of-sync case, please see #112 

>  however existing changes will most likely upgrade to beefy pallet,

I think I don't get, what you mean by this. Could you please elaborate?

tomusdrw: > I think I don't get, what you mean by this. Could you please elaborate?

What I mean is that BEEFY will be introduced as a runtime upgrade to chains like Kusama, Polkadot and maybe even Rococo.
This means that the node (and the gadget code) needs to handle a case where the pallet is not present gracefuly. The client will be upgraded first and after that will the runtime get the pallet. This means that the gadget needs to be able to enable itself when it detect that the pallet becomes active in the runtime.

adoerr: Thanks for the clarification! Filed #113 for this.

PR #114, opened 2021-03-10
Bump substrate.

tomusdrw:


adoerr (approved):
LGTM. 

Clippy currently fails due to an outdated version. `clamp()` is stable since 1.50

PR #115, opened 2021-03-10
Update toolchain

adoerr:
Clippy for #114 fails due to an outdated version. We should be able to just use nightly.


adoerr:
Had to downgrade clippy to to stable due to some issue with `construct_runtime!()`.

adoerr:
Guess, I went full-circle on this one :-)

tomusdrw (approved):

PR #117, opened 2021-03-15
Bump Substrate and Deps

adoerr:
- Bump Substrate again in order to make progress on #116 
- Remove `new_light()` since we use the full node only anyways and updating the light client part becomes tedious.

### DIFF
--- a/beefy-node/node/src/service.rs
+++ b/beefy-node/node/src/service.rs
@@ -1,16 +1,14 @@
 //! Service and ServiceFactory implementation. Specialized wrapper over substrate service.
-use {std::sync::Arc, std::time::Duration};
+use std::{sync::Arc, time::Duration};
 
-use {
-	beefy_node_runtime::{self, opaque::Block, RuntimeApi},
-	sc_client_api::ExecutorProvider,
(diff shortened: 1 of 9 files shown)


beefy-node/node/src/service.rs
```diff
@@ -1,18 +1,16 @@
...
-use sc_executor::native_executor_instance;
-use sc_finality_grandpa::SharedVoterState;
-use sc_service::{error::Error as ServiceError, Configuration, TaskManager};
-use sc_telemetry::TelemetrySpan;
-
-use sp_consensus_aura::sr25519::AuthorityPair as AuraPair;
-use sp_inherents::InherentDataProviders;
-
-use beefy_node_runtime::{self, opaque::Block, RuntimeApi};
+use {std::sync::Arc, std::time::Duration};
```
tomusdrw, line 2: Can we use rustfmt's `--check --config merge_imports=true` to have the formatter enforce uniform imports?

See [example here](https://github.com/tomusdrw/rust-web3/blob/master/.github/workflows/lint.yml#L35).

tomusdrw (approved):

PR #118, opened 2021-03-16
Start BEEFY test support lib

adoerr:
Introduce a BEEFY test support library. 

The idea for this lib is to bundle all the test related functionality, so that it can be referenced as one dev-dependency. Also the light client test implementation should be moved. here.


beefy-test/src/keyring.rs
```diff
@@ -0,0 +1,111 @@
...
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
```
tomusdrw, line 16:
```suggestion

#![warn(missing_docs)]

```

beefy-test/src/keyring.rs
```diff
@@ -0,0 +1,111 @@
...
+		format!("//{}", self)
+	}
+
+	/// Iterator over all test accounts
+	pub fn iter() -> impl (Iterator<Item = Keyring>) {
+		<Self as strum::IntoEnumIterator>::iter()
+	}
+}
+
+impl From<Keyring> for &'static str {
```
tomusdrw, line 65: Why is that useful / needed?

adoerr: It is used for `Keyring::pair()`

tomusdrw: But we also have `to_seed` and it's using `strum::Display`? Why do we really need both and risk getting mismatch?

adoerr: Changed the `Keyring::pair()` implementation.

tomusdrw (approved):

Issue #113, opened 2021-03-09
Beefy gadget/client should allow for delayed Beefy pallet initialization


adoerr:
From #95 
> The assumption here is that the beefy pallet is part of the runtime (pretty much since genesis), however existing changes will most likely upgrade to beefy pallet, so I think we should be handling this error a bit more gracefuly.

> What I mean is that BEEFY will be introduced as a runtime upgrade to chains like Kusama, Polkadot and maybe even Rococo.
This means that the node (and the gadget code) needs to handle a case where the pallet is not present gracefuly. The client will be upgraded first and after that will the runtime get the pallet. This means that the gadget needs to be able to enable itself when it detect that the pallet becomes active in the runtime.

PR #121, opened 2021-03-22
Delayed BEEFY worker initialization

adoerr:
Resolves: #113

### DIFF
--- a/beefy-gadget/src/lib.rs
+++ b/beefy-gadget/src/lib.rs
@@ -14,20 +14,23 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+use std::{convert::TryFrom, fmt::Debug, sync::Arc};
+
 use beefy_primitives::BeefyApi;
 use codec::Codec;
-use sc_client_api::{Backend as BackendT, BlockchainEvents, Finalizer};
+
+use sc_client_api::{Backend, BlockchainEvents, Finalizer};
 use sc_network_gossip::{
 	GossipEngine, Network as GossipNetwork, ValidationResult as GossipValidationResult, Validator as GossipValidator,
 	ValidatorContext as GossipValidatorContext,
 };
+
 use sp_api::{BlockId, ProvideRuntimeApi};
 use sp_application_crypto::AppPublic;
 use sp_blockchain::HeaderBackend;
 use sp_consensus::SyncOracle as SyncOracleT;
 use sp_keystore::SyncCryptoStorePtr;
-use sp_runtime::traits::{Block as BlockT, Zero};
-use std::{convert::TryFrom, fmt::Debug, sync::Arc};
+use sp_runtime::traits::{Block, Zero};
 
 mod error;
 mod round;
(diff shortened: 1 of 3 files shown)


beefy-gadget/src/lib.rs
```diff
@@ -52,74 +55,89 @@ pub fn beefy_peers_set_config() -> sc_network::config::NonDefaultSetConfig {
...
 	}
 }
 
-pub async fn start_beefy_gadget<Block, Pair, Backend, Client, Network, SyncOracle>(
-	client: Arc<Client>,
+/// Start the BEEFY gadget.
+///
+/// This is a thin shim around running and awaiting a BEEFY worker. The [`Client`]
+/// convenience trait is not used here on purpose. We don't want to leak it into the
+/// public interface of the BEEFY gadget.
```
tomusdrw, line 109: > We don't want to leak it into the public interface of the BEEFY gadget.

Any compelling reason why we we don't want to expose it?

beefy-gadget/src/lib.rs
```diff
@@ -52,74 +55,89 @@ pub fn beefy_peers_set_config() -> sc_network::config::NonDefaultSetConfig {
...
+pub async fn start_beefy_gadget<B, P, BE, C, N, SO>(
```
tomusdrw, line 110: I'm not a huge fan of this one-letter generic parameters, it's so easy to forget what they are supposed to represent. Personally I like the (java?) `TBlock`, `TPair`, etc convention where `T` makes it clear it's a generic parameter, but the rest of the name gives you some clue what it actually is.

beefy-gadget/src/worker.rs
```diff
@@ -44,76 +61,136 @@ struct VoteMessage<Hash, Number, Id, Signature> {
...
+	C: Client<B, BE, P>,
+	C::Api: BeefyApi<B, P::Public>,
 {
+	/// Retrun a new BEEFY worker instance.
+	///
+	/// Note that full BEEFY worker initialization can only be completed, if an
+	/// on-chain BEEFY pallet is available. Reason is that the current active
+	/// validator set has to be fetched from the on-chain BEFFY pallet.
+	///
+	/// For this reason, BEEFY worker initialization complets only after a finality
```
tomusdrw, line 116:
```suggestion
	/// For this reason, BEEFY worker initialization completes only after a finality
```

beefy-gadget/src/worker.rs
```diff
@@ -244,12 +321,21 @@ where
 			futures::select! {
 				notification = self.finality_notifications.next().fuse() => {
 					if let Some(notification) = notification {
+						if self.state == State::New {
+							match self.init_validator_set() {
+								Ok(()) => (),
+								Err(err) => {
+									error!(target: "beefy", "🥩 Init validator set failed: {:?}", err);
+									return;
```
tomusdrw, line 329: :thinking: this shouldn't be an error and it should not really stop the entire worker, should it?

beefy-gadget/src/lib.rs
adoerr: Main reason, why I didn't want to expose it is that it would require to change every node which is using the BEEFY gadget. Also `Client` could be seen as a minimal bound and a client is actually free to pass a more 'rich' client.

beefy-gadget/src/lib.rs
adoerr: This is the first case, we really differ in preferences, I guess. For me a generic parameter is, well generic, i.e. meaningless. The meaning or semantic is induced be the where clause. I also don't like to duplicate information which is easily provided by common tools like rust-analyzer.

adoerr: Question is, how we want to deal with those kind of preferences, both as a team and as a company. Establish and enforce a style guide or accept different preferences. As always, both options have pros and cons.

beefy-gadget/src/worker.rs
adoerr: I'm not sure. It should not cause any harm to turn the error into a warning. The worker state will remain `State::New` and no voting will be done.

beefy-gadget/src/lib.rs
tomusdrw: That's fair, although since the trait has blanket-impl, would that really be an issue? Like no one would really manually implement that trait anyway, no?

beefy-gadget/src/lib.rs
tomusdrw: We don't have a formal framework how to deal with that, but I'm not strong enough on my preference to keep defending it :) I can work in both, and prefer consistency over any specific choice. Since it's mainly you working on that repo, I think you should be the one making the decision.

beefy-gadget/src/lib.rs
adoerr: I will turn that into an issue which should be a good one get done once things stabilize a bit more.

beefy-gadget/src/lib.rs
adoerr: Appreciate that!

beefy-gadget/src/worker.rs
tomusdrw: Yeah, I think this should at most be a `warn`. I say "at most", because this is not really actionable by the node operator, they can only ignore this and wait for the runtime to include the pallet and in case of the transition period this will definitely be triggered.

Also if we expect the worker to remain in `State::new` we shouldn't be doing `return` here, should we?

adoerr: Will turn it into an info and continue.

beefy-gadget/src/lib.rs
adoerr: See #122

tomusdrw (approved):

PR #124, opened 2021-03-23
Lower log-level for a missing validator set

adoerr:
I wounder, if [best_finalized_block](https://github.com/paritytech/grandpa-bridge-gadget/blob/9d407131f54f38343b92adebb6929baa2f38ce83/beefy-gadget/src/worker.rs#L134) should actually be initialized in `init_validator_set()` instead of `BeefyWorker::new()`.

Otherwise, if a BEEFY enabled client runs for a substantial amount if time, before the BEEFY pallet gets deployed on-chain we may end up voting for an old block for which we don't even have the header. Reason is that currently it is assumed that  the block to vote on is not more than (`best_finalized_block `/ 2)`.next_power_of_two()` away.

### DIFF
--- a/beefy-gadget/src/worker.rs
+++ b/beefy-gadget/src/worker.rs
@@ -115,7 +115,7 @@ where
 	///
 	/// For this reason, BEEFY worker initialization completes only after a finality
 	/// notification has been received. Such a notifcation is basically an indication
-	/// that an on-chain BEEFY pallet is available.
+	/// that an on-chain BEEFY pallet may be available.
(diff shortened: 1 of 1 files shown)


tomusdrw (approved):
> I wounder, if best_finalized_block should actually be initialized in init_validator_set() instead of BeefyWorker::new().

Yes, I think that's reasonable. It should basically be `best_finalized_block_with_beefy_pallet`, cause without beefy pallet it doesn't make any sense to start voting on anything.

Issue #122, opened 2021-03-22
Make BEEFY gadget `Client` trait public


adoerr:
Make the BEEFY gadget `Client` trait public and adjust client code accordingly.

PR #126, opened 2021-03-24
Make Client convenience trait public

adoerr:
Resolves: #122 

Since the public interface changes because of the Prometheus stuff anyway, let's get this done as well.


beefy-gadget/src/lib.rs
```diff
@@ -123,7 +121,7 @@ pub async fn start_beefy_gadget<B, P, BE, C, N, SO>(
 	P::Public: AppPublic + Codec,
 	P::Signature: Clone + Codec + Debug + PartialEq + TryFrom<Vec<u8>>,
```
tomusdrw, line 122: Both `P::Public` and `P::Signature` bounds are redundant as well now, no?

tomusdrw (approved):

beefy-gadget/src/lib.rs
adoerr: Don't thing so, `P::Public` is used as a bound for `BeefyApi` and `P::Signature` is a bound on `BeefyWorker`. I think, might could do with one less bound on `BeefyWorker`. Need to look into that later.

Going to merge the PR for now

PR #132, opened 2021-03-30
Bump Substrate and Deps

adoerr:
Keep up with Substrate.


tomusdrw (approved):

tomusdrw:
BTW, can we maybe automate that via dependbot? I'd love to have dependabot submit a PR like that once a week.

PR #137, opened 2021-04-02
Use lock only for CI.

tomusdrw:
Having a `Cargo.lock` present in a library might cause issues in polkadot repo (duplicated/locked dependencies).
This PR attempts to remove `Cargo.lock` file, while still maintaining some predictability (i.e. not falling back to latest versions always).

The idea is that `Cargo.lock.ci` file is present and whenever CI is executed it's renamed to `Cargo.lock` file. Locally, when the project is checked out, the developer should also copy the `.ci` file and name it `Cargo.lock`. When bumping dependencies the opposite is required.

Note I'm not 100% sure it would solve polkadot problems, but I think it's worth trying.


adoerr:
Given that the BEEFY `Cargo.lock` will not influence any project which uses BEEFY as a dependency, I'm a bit hesitant to introduce some 'special' workflow.

AFIU, as long as we keep BEEFY close to Substrate master there should not be a problem with the Polkadot repo.

adoerr (approved):
Change as such LGTM.

But I'm not sure we should do this. I'm especially  a bit worried about the 'manual copy' part.

tomusdrw:
I was wrong how `Cargo.lock` might influence dependencies, so let's close this as it doesn't provide any value.

PR #139, opened 2021-04-05
beefy: authority set changes fixes

### LINKED ISSUE #135: Track best BEEFY finalized block
Currently, BEEFY tracks the best GRANDPA finalized block in `best_finalized_block`. Instead what we need to track is the best block 'finalized' by BEEFY. That would be the `block_number` in the most recent BEEFY signed commitment. 

- For a BEEFY gadget in `State::New`, wait for a `SignedCommitment` notification received through gossip.
- Update `best_finalized_block` with `SignedCommitment.block_number`.
- For each `SignedCommitment` notification received, update `best_finalized_block` again, if `SignedCommitment.block_number` is more recent.

Gossip will periodically send out `SignedCommitments` to allow for initial signed commitment discovery (first item in the list above).

andresilva:
Pratically fixes #135 (although there's more stuff in there). Changes the validator set logic to always apply changes regardless of whether we should vote on a given block or not.

### DIFF
--- a/beefy-gadget/src/worker.rs
+++ b/beefy-gadget/src/worker.rs
@@ -295,8 +295,6 @@ where
 				(message.id, message.signature),
 			);
 		}
-
-		self.best_finalized_block = *notification.header.number();
 	}
 
 	fn handle_vote(&mut self, round: (MmrRootHash, NumberFor<B>), vote: (P::Public, S)) {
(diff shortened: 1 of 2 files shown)


adoerr (approved):
LGTM

tomusdrw (approved):

PR #140, opened 2021-04-05
gadget: remove superfluous signature type

andresilva:
Merge after #139.

Remove the superfluous `Signature` type from the gadget since it should be derived from the `Pair`.


adoerr (approved):
LGTM

tomusdrw (approved):

PR #141, opened 2021-04-05
gadget: reduce gossip spam

andresilva:
Merge after #140.
 
Keep track of the last `N` live rounds and only allow gossiping messages for those rounds.


beefy-gadget/src/worker.rs
```diff
@@ -55,12 +63,101 @@ where
...
+		if VoteMessage::<MmrRootHash, NumberFor<B>, P::Public, P::Signature>::decode(&mut data).is_ok() {
+			GossipValidationResult::ProcessAndKeep(self.topic)
+		} else {
+			GossipValidationResult::Discard
+		}
+	}
+
+	fn message_expired<'a>(&'a self) -> Box<dyn FnMut(B::Hash, &[u8]) -> bool + 'a> {
+		let live_rounds = self.live_rounds.read();
+		Box::new(move |_topic, mut data| {
```
adoerr, line 130: `topic` is ignored here, because we only have the single global 'beefy' topic anyway, right?

beefy-gadget/src/worker.rs
```diff
@@ -55,12 +63,101 @@ where
...
+
+		// NOTE: ideally we'd use a VecDeque here, but currently binary search is only available on
+		// nightly for `VecDeque`.
+		while live_rounds.len() > MAX_LIVE_GOSSIP_ROUNDS {
+			let _ = live_rounds.remove(0);
+		}
+
+		if let Some(idx) = live_rounds.binary_search(&round).err() {
+			live_rounds.insert(idx, round);
+		}
```
adoerr, line 100: The sorted insert is crucial to make this work, because validators can/will see vote messages in different order, right?

beefy-gadget/src/worker.rs
```diff
@@ -21,20 +21,22 @@ use std::{
...
 
-use beefy_primitives::{
-	BeefyApi, Commitment, ConsensusLog, MmrRootHash, SignedCommitment, ValidatorSet, BEEFY_ENGINE_ID, KEY_TYPE,
-};
 use codec::{Codec, Decode, Encode};
 use futures::{future, FutureExt, StreamExt};
 use hex::ToHex;
 use log::{debug, error, info, trace, warn};
-use parking_lot::Mutex;
+use parking_lot::{Mutex, RwLock};
```
adoerr, line 28: What is actually the advantage of these versions over the `std::sync` ones?

adoerr (approved):
Couple of questions for the benefit of my understanding. 

Other than that, LGTM.

beefy-gadget/src/worker.rs
andresilva: Yep. In the future once we have one topic per round we might be able to just look at the topic to figure out if a message is expired or not.

beefy-gadget/src/worker.rs
andresilva: Yeah, my thinking was that currently we just start a round whenever we a get message for it, and this could come out-of-order, in which case we could end up pruning rounds that aren't the oldest ones.

beefy-gadget/src/worker.rs
andresilva: tldr: they are faster

There is some useful information here: https://github.com/Amanieu/parking_lot/blob/master/README.md https://webkit.org/blog/6161/locking-in-webkit/
It's likely that in the future the locks in the standard library will just be replaced by this implementation.

beefy-gadget/src/worker.rs
```diff
@@ -55,12 +63,101 @@ where
 	<<B::Header as Header>::Hashing as Hash>::hash(b"beefy")
 }
 
+/// Allows messages from last `MAX_LIVE_GOSSIP_ROUNDS` to flow, everything else gets
```
tomusdrw, line 66:
```suggestion
/// Allows messages from last [`MAX_LIVE_GOSSIP_ROUNDS`] to flow, everything else gets
```

beefy-gadget/src/worker.rs
```diff
@@ -55,12 +63,101 @@ where
...
+		let mut live_rounds = self.live_rounds.write();
+
+		// NOTE: ideally we'd use a VecDeque here, but currently binary search is only available on
+		// nightly for `VecDeque`.
+		while live_rounds.len() > MAX_LIVE_GOSSIP_ROUNDS {
+			let _ = live_rounds.remove(0);
+		}
+
+		if let Some(idx) = live_rounds.binary_search(&round).err() {
+			live_rounds.insert(idx, round);
```
tomusdrw, line 99: Wouldn't it make sense to use `BTreeSet` instead? Pruning can easily be done with `split_off`.

That said with 5 elements it probably doesn't matter at all performance wise even if we did linear search, but I'd consider it if it makes the code simpler.

beefy-gadget/src/worker.rs
```diff
@@ -280,26 +384,28 @@ where
...
-		self.best_finalized_block = *notification.header.number();
+			self.gossip_engine
+				.lock()
+				.gossip_message(topic::<B>(), encoded_message, false);
+		}
 	}
 
-	fn handle_vote(&mut self, round: (MmrRootHash, NumberFor<B>), vote: (P::Public, S)) {
+	fn handle_vote(&mut self, round: (MmrRootHash, NumberFor<B>), vote: (P::Public, P::Signature)) {
+		self.gossip_validator.note_round(round.1);
```
tomusdrw, line 407: Two things  that pop into my mind:
1. `handle_vote` is being called in 2 places (when we send the vote and when we receive vote from gossip). IMHO the `note_round` should only happen in the first case. It's safe now, cause we use it in the gossip validator anyway, but still I think it would be cleaner.
2. Non-voting workers will now never gossip anything. We should probably change the logic of `handle_finality_notification`, so that in case `should_vote_on` is `true` with `note_round` unconditionally.

beefy-gadget/src/worker.rs
```diff
@@ -55,12 +63,101 @@ where
...
+	P::Public: Decode,
+	P::Signature: Decode,
+{
+	fn validate(
+		&self,
+		_context: &mut dyn GossipValidatorContext<B>,
+		_sender: &sc_network::PeerId,
+		mut data: &[u8],
+	) -> GossipValidationResult<B::Hash> {
+		if VoteMessage::<MmrRootHash, NumberFor<B>, P::Public, P::Signature>::decode(&mut data).is_ok() {
```
tomusdrw, line 121: I suppose this happens after `message_allowed` returns true, right? Shouldn't we actually verify signatures here already?

tomusdrw (commented):
looks good, but I'd like to make sure that non-validator nodes propagate as well.

beefy-gadget/src/worker.rs
andresilva: Currently `BTreeSet` does not provide any methods for poping or looking at the head. I could inspect the contents of the `BTreeSet` with an iterator but then we're back to linear search. What was your idea of using `split_off`? If the set has elements `[1,2,4,5]` and I want to insert `3` I don't actually want to prune anything since it only has 4 elements.

beefy-gadget/src/worker.rs
andresilva: This is the method that's called first by the gossip layer when we receive the message for the first time (and only called once while we still hold on to this message). So yes, this would be a good place to verify signatures as we can immediately discard the message if the signature is invalid. We are currently not checking signatures though AFAIK.

beefy-gadget/src/worker.rs
tomusdrw: I was assuming that eventually the rounds will be consecutive anyway (i.e. we will eventually receive all of them, but maybe in some scrumbled order). So instead of storing up to 5 entries, I would go for storing at most entries from rounds `[N-5; N]`. Hence `split_off` could simply be used to get rid of all old rounds. The inspection would not really be needed if we just stored the best session we've seen so far.

beefy-gadget/src/worker.rs
tomusdrw: > They should be as they call handle_vote whenever they get a gossip message, and this will note the round in the gossip validator.

So that what confuses me. The `GossipValidator` seems to reject messages that have not been `noted` yet, and the only place where we `note_round` is here, so it will only be noted if the node is validating.

beefy-gadget/src/worker.rs
andresilva: > I was assuming that eventually the rounds will be consecutive anyway

Yeah, I think so but I'm not sure it is completely guaranteed with the current logic. The `N` here will be some random block number that fits the power of two rule so we can't easily figure out what the range of values in that vec should be.

beefy-gadget/src/worker.rs
andresilva: The entry point for the gossip validator is the `validate` method, which only tries to decode the message. Returning `GossipValidationResult::ProcessAndKeep` will make sure the beefy listener will receive the message. `message_allowed` and `message_expired` are called later on when deciding whether to send a particular message to a given peer, or whether to prune the message from the gossip layer. Those methods will only be called after the beefy worker has processed the message and its round should already have been noted by then.

tomusdrw: Ah, that makes sense, if `message_allowed` is only called for outgoing messages we are good!

tomusdrw (approved):

Issue #100, opened 2021-03-05
Validate signatures for votes received from peers


adoerr:
`BeefyWorker::handle_vote()` should validate vote signatures before processing them.

andresilva:
#142 implemented verification of the signatures but we still need to make sure the given signer is part of the validator set.

adoerr:
Verify , that the `msg.id` in [this](https://github.com/paritytech/grandpa-bridge-gadget/blob/9be54d3e816d8e8e7aae392f70d5c6531ad00b26/beefy-gadget/src/worker.rs#L122) line is a member of the validator set.

tomusdrw:
This could also be done in `Round` structure (`add_vote`) or earlier.

Issue #17, opened 2020-11-02
BEEFY justifications (SignedCommitments) should be stored in the DB


tomusdrw:
Storing them as justifications will expose them via `getBlock` RPC - we return all kinds of justifications we have for the block (BABE, GRANDPA, BEEFY, etc).

PR #142, opened 2021-04-08
gadget: verify SignedCommitment message signature

andresilva:


beefy-gadget/src/worker.rs
```diff
@@ -109,20 +109,25 @@ impl<B, P> GossipValidator<B> for BeefyGossipValidator<B, P>
...
+		sender: &sc_network::PeerId,
 		mut data: &[u8],
 	) -> GossipValidationResult<B::Hash> {
-		if VoteMessage::<MmrRootHash, NumberFor<B>, P::Public, P::Signature>::decode(&mut data).is_ok() {
-			GossipValidationResult::ProcessAndKeep(self.topic)
-		} else {
-			GossipValidationResult::Discard
+		if let Ok(msg) = VoteMessage::<MmrRootHash, NumberFor<B>, P::Public, P::Signature>::decode(&mut data) {
+			if P::verify(&msg.signature, &msg.commitment.encode(), &msg.id) {
+				return GossipValidationResult::ProcessAndKeep(self.topic);
```
tomusdrw, line 123: I guess that's good enough, but long-term we should also make sure that `msg.id` is part of the declared validator set id. That needs persistence of previous validator sets though (#17).

tomusdrw (approved):

andresilva:
@tomusdrw I'll comment on #100 and keep it open.

adoerr (approved):
LGTM

Issue #112, opened 2021-03-09
Initialize BeefyWorker validator set id for out-of-sync BEEFY gadget/client


adoerr:
From #95 

> So this obviously works for now, but what if the client is not fully in sync with the network? I think we need a bit more sophisticated mechanism, where we actually update the validator set id when new blocks are imported / we get synced up.

Currently the `BeefyWorker` validator set id is always initialized by using `BeefyApi::validator_set()`. This assumes the client is synced with the network and that the `beefy_pallet` session handler has seen all relevant `on_new_sesseion()` calls.

As per the above comment, this approach is not sufficient for an out-of-sync client.

tomusdrw:
Related #183
We need to be notified by the block import pipeline about already seen justifications and defer starting the gadget when major sync is on-going.

seunlanlege:
can be closed now that  paritytech/substrate#10882 is merged

acatangiu:
The issue description is a bit outdated so I will provide an update here.

With Lean BEEFY a node coming online at session N doesn't currently have a way to sync older mandatory BEEFY blocks to get the uninterrupted authority hand-offs chain. Instead, when mandatory block N is BEEFY-finalized, it assumes the finalizing super-majority of validators is honest. The node starts voting in session N+1 based on the finalized mandatory BEEFY block N authority set hand-off.

This issue tracks a way for the node to start voting - in lean beefy just wait for gossip of some mandatory block justification, then start voting on next session based on it - which is implemented.

Support for having the node able to vote as soon as sync is done is tracked in https://github.com/paritytech/grandpa-bridge-gadget/issues/183

So yes, we can mark this issue fixed by https://github.com/paritytech/substrate/pull/10882.

PR #144, opened 2021-04-09
Milestone 1

adoerr:
Changes we have identified for `Milestone 1`:

- `init_validator_set()` should use `best_finalized` and not `best_hash`
- `client.info()` should be stored to prevent races
- `best_finalized_block` should be optional
- `should_vote()` should bail if `best_finalized_block` is  `None` 
- get rid of worker state
- handle transition from 'no-pallet' to 'pallet' and vice versa

That should cover a non-optimized happy case.


beefy-gadget/src/worker.rs
```diff
@@ -304,7 +305,13 @@ where
 			return false;
 		}
 
-		let diff = self.best_finalized_block.saturating_sub(self.best_block_voted_on);
+		let best_finalized = if let Some(block) = self.best_finalized_block {
+			block
+		} else {
+			return false;
```
tomusdrw, line 311: I'd add `debug/trace` logs in every `return false` so that we can easily figure out why some validator is not voting.

adoerr: Done

tomusdrw (approved):

adoerr:
Resolves: #112 

Re-syncing of a BEEFY node is done by calling [validator_set()](https://github.com/paritytech/grandpa-bridge-gadget/blob/1ea7f45c87a3e887d2ee8a00c80e1d25ad823141/beefy-gadget/src/worker.rs#L292-L299) as part of finality notification processing.

If a BEEFY node is out-of-sync, upon receiving a finality notification, it will either

- use the validator set enacted through an `AuthoritiesChange` digest
- fetch the validator set from BEEFY on-chain state

beefy-gadget/src/worker.rs
```diff
@@ -334,26 +283,59 @@ where
 		Ok(sig)
 	}
 
+	/// Return the current active validator set.
+	///
+	/// Note that the validator set could be `None`. This is the case if we don't find
+	/// a BEEFY authority set change and we can't fetch the validator set from the
+	/// BEEFY on-chain state. Such a failure is usually an indication that the BEEFT
+	/// pallet has not been deplyed (yet).
```
tomusdrw, line 291:
```suggestion
	/// pallet has not been deployed (yet).
```

beefy-gadget/src/worker.rs
```diff
@@ -334,26 +283,59 @@ where
...
+	fn validator_set(&self, header: &B::Header) -> Option<ValidatorSet<P::Public>> {
+		if let Some(new) = find_authorities_change::<B, P::Public>(header) {
+			Some(new)
+		} else {
+			let at = BlockId::hash(self.client.info().finalized_hash);
```
tomusdrw, line 296:
```suggestion
			let at = BlockId::hash(header.hash());
```

I think this will make for much more predictable behavior.

beefy-gadget/src/worker.rs
```diff
@@ -334,26 +283,59 @@ where
 		Ok(sig)
 	}
 
+	/// Return the current active validator set.
```
tomusdrw, line 286:
```suggestion
	/// Return the current active validator set at given header.
```

beefy-gadget/src/worker.rs
```diff
@@ -334,26 +283,59 @@ where
...
+			if (active.id != self.rounds.validator_set_id()) || (active.id == GENESIS_AUTHORITY_SET_ID) {
+				self.rounds = round::Rounds::new(active.clone());
+				debug!(target: "beefy", "🥩 New Rounds for id: {:?}", active.id);
+			}
 		};
 
+		// NOTE: currently we act as if this block has been finalized by BEEFY as we perform
+		// the validator set changes instantly (insecure). Once proper validator set changes
+		// are implemented this should be removed
+		self.best_finalized_block = Some(*notification.header.number());
```
tomusdrw, line 335: I think this is a little bit confusing and I think we still miss a bit of logic. Let me elaborate.

We have two block variables that both affect `should_vote_on`:
1. `best_finalized_block`
2. `best_block_voted_on`

I think we kept confusing the two and probably the implementation instructions were incorrect.

So the `should_vote_on` formula should be roughly something like that:
```
(best_observed_block - best_beefy_finalized_block).next_power_of_two()
```

The logic for current variables:
1. `best_finalized_block` = `best_observed_block`, but with extra `None` value right after starting up.
2. `best_block_voted_on` = in case the node is voting, this will get updated with last vote.

I think we need to change the logic (and maybe the name) of `best_block_voted_on` instead of `best_finalized_block`. Basically:
1. `best_finalized_block` should simply be updated every time we receive GRANDPA finality notification (IMHO we should even call it `best_grandpa_block`).
2. `best_beefy_block` should be updated in every case from this list:
    1. We conclude a round (i.e there exists `SignedCommitment`).
    2. We transition to new validator set.
   
The second case is obviously a temporary measure for `Milestone 1`.

tomusdrw (approved):

PR #149, opened 2021-04-15
Remove `branch = "master"` from toml files 

tomusdrw:
To allow referencing custom branches via `revision` in `Cargo.lock`.


adoerr (approved):
LGTM.

Still would want to wait with a merge, until we have some sort of idea how we are going to deal with Substrate API breaking changes.

tomusdrw:
Closing, since we need more sustainable solution.

PR #150, opened 2021-04-15
Use wildcard for dep patch grouping

adoerr:
Use wildcard to specify external Cargo dependencies. Specifically for the `patch` grouping. This allows for easier integration with external repositories, like Bridges.

Using Wildcards will also assure, that **only** the patch level version is upgraded, regardless of how a version string looks like.

### DIFF
--- a/beefy-gadget/rpc/Cargo.toml
+++ b/beefy-gadget/rpc/Cargo.toml
@@ -6,15 +6,15 @@ edition = "2018"
 license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
 
 [dependencies]
-futures = { version = "0.3.14", features = ["compat"] }
-log = "0.4"
-serde = { version = "1.0.125", features = ["derive"] }
-serde_json = "1.0.64"
+futures = { version = "0.3.*", features = ["compat"] }
+log = "0.4.*"
+serde = { version = "1.0.*", features = ["derive"] }
+serde_json = "1.0.*"
 
-jsonrpc-core = "15.1.0"
-jsonrpc-core-client = "15.1.0"
-jsonrpc-derive = "15.1.0"
-jsonrpc-pubsub = "15.1.0"
+jsonrpc-core = "15.1.*"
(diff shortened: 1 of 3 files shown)


tomusdrw (commented):
I don't think it's the right choice. 

> The string "0.1.12" is a semver version requirement. Since this string does not have any operators in it, it is interpreted the same way as if we had specified "^0.1.12", which is called a caret requirement.

This means that our versions are minimal versions that we expect from downstream applications.

If we change this to wildcard, we risk that the downstream project would not compile or work incorrectly because it would attempt to use version which is older than what we expect.

IMHO we should keep full semver versions, but make sure they stay as low as possible instead.

adoerr:
Well, I ran immediately into issues integrating BEEFY with Millau, because Millau is using a `low` version of `futures`.
 
I personally would pivot towards the most `recent` version possible, not the oldest one.

adoerr:
Also, if you have a version like "1.0.12", it could update up to "1.9.99", no? In other words, Caret dependencies behave differently,, based on the version string.

tomusdrw:
> Well, I ran immediately into issues integrating BEEFY with Millau, because Millau is using a low version of futures.

Can you share the exact details of the issue? If Millau requires say `1.0.5` and BEEFY requires `1.0.12` then cargo should resolve the lowest matching version to `1.0.12` (effectively bumping what Millau is using), unless the version is pinned (via `~`). If there are some changes that prevent Millau from working with `1.0.12` then I'd say we should consider lowering the version requirement in BEEFY?

> I personally would pivot towards the most recent version possible, not the oldest one.

But that's completely opposite to what you did in this PR? Also BEEFY is not a product on it's own, so the goal is to make the runtime developer's lives as easy as possible - I don't think we are in position (or even should) set vaguely justified requirements for downstream crates.

> Also, if you have a version like "1.0.12", it could update up to "1.9.99", no? 

Yes, that's correct. But since minor versions are supposed to be backward compatible according to semver then I think it's expected as well, no?

Issue #119, opened 2021-03-18
Expose BEEFY prometheus metrics.


tomusdrw:
After a "production" deployment of BEEFY (for instance on Westend / Rococo) it will be really nice to collect some metrics and get some insight of what happens inside the BEEFY gadget / gossip.
I'd see what metrics we expose for GRANDPA and copy that, but what would be interesting from the top of my head:
1. Current Validator Set Id
2. Round state: starting / concluding
3. Number of valid signed commitments received per round (with required threshold information)
4. Number of incoming / outgoing gossip messages / commitments per round.

adoerr:
Prometheus metric setup has been done with #125. Current validator set id is available as a metric as part of this PR as well.

Further metrics need a bit of a refinement, though. Exposing anything other than a simple numeric value is not really straightforward. So a metric is basically either a value or a counter.

PR #152, opened 2021-04-16
Add additional metrics

adoerr:
Resolves: #119 

We now have metrics for:

- current active validator set id
- total number of votes sent by BEEFY node
- most recent concluded voting round
- best blocked finalized by BEEY
- next block BEEFY should vote on
- number of sessions without a signed commitment (aka skipped sessions)

I have tested those metrics against a local Prometheus server.

### DIFF
--- a/beefy-gadget/src/worker.rs
+++ b/beefy-gadget/src/worker.rs
@@ -46,6 +46,7 @@ use beefy_primitives::{
 
 use crate::{
 	error::{self},
+	metric_inc, metric_set,
 	metrics::Metrics,
 	notification, round,
 	validator::{topic, BeefyGossipValidator},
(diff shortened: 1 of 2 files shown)


beefy-gadget/src/worker.rs
```diff
@@ -227,6 +238,8 @@ where
 				debug!(target: "beefy", "🥩 New Rounds for id: {:?}", active.id);
 
 				self.best_beefy_block = Some(*notification.header.number());
+
+				metric_set!(self, beefy_best_block, *notification.header.number());
```
tomusdrw, line 242: Depending on what we want the metric exactly to be we might or might not want to have this line. This is kind of a "forced BEEFY change", cause the block is not really beefy-finalized (as metric decsription suggests). I think it's fine for now, since we want to have insight about internal state, but would add a comment here to make sure we get rid of this later on.

tomusdrw (approved):
lgtm

beefy-gadget/src/worker.rs
adoerr: Yeah, getting rid of this one was the idea. Will clarify.

andresilva (commented):
Too late, but lgtm.

PR #155, opened 2021-04-19
Bump substrate

andresilva:
Revert #154 and update substrate using `cargo update -p sp-io` instead of updating all crates.


adoerr (approved):
LGTM

PR #157, opened 2021-04-20
don't log error on missing validator keys

andresilva:
This isn't an error, this message will get printed for all non-validators.


andresilva:
Sorry for not fixing these all at once but I'm fixing stuff as I see it.

tomusdrw (approved):
looks good! Can you please fix the warning as well?

PR #158, opened 2021-04-21
Fix validator set change handling

andresilva:


beefy-gadget/src/worker.rs
```diff
@@ -212,27 +212,28 @@ where
...
-
 			// Authority set change or genesis set id triggers new voting rounds
 			//
 			// TODO: (adoerr) Enacting a new authority set will also implicitly 'conclude'
 			// the currently active BEEFY voting round by starting a new one. This is
-			// temporary and needs to be repalced by proper round life cycle handling.
-			if (active.id != self.rounds.validator_set_id()) || (active.id == GENESIS_AUTHORITY_SET_ID) {
+			// temporary and needs to be replaced by proper round life cycle handling.
+			if active.id != self.rounds.validator_set_id()
+				|| (active.id == GENESIS_AUTHORITY_SET_ID && self.best_beefy_block.is_none())
```
andresilva, line 227: On genesis we only do this once instead of for all finalized blocks as we were doing previously.

beefy-gadget/src/worker.rs
```diff
@@ -212,27 +212,28 @@ where
...
+			{
+				debug!(target: "beefy", "🥩 New active validator set id: {:?}", active);
+				metric_set!(self, beefy_validator_set_id, active.id);
+
+				// BEEFY should produce a signed commitment for each session
+				if active.id != self.last_signed_id + 1 && active.id != GENESIS_AUTHORITY_SET_ID {
```
andresilva, line 233: Unlike before this is only done when we actually change the validator set id.

adoerr (approved):
LGTM

tomusdrw (approved):

PR #160, opened 2021-04-23
Fix should_vote_on()

adoerr:
Use `best_beefy_block` to calculate next block to vote on instead of `best_block_voted_on`.

This fixes an issue if a chain gets restarted. Let the best GRANDPA finalized block after a restart be `1024`. Due to the restart, `best_block_voted_on` will be initialized to `0`. In this case `should_vote_on()` will calculate `next_block_to_vote_on` as `512`. In other words, we would never 'catch up' in terms of which block number to vote on, because `best_block_voted_on` stays at `0`

### DIFF
--- a/beefy-gadget/src/worker.rs
+++ b/beefy-gadget/src/worker.rs
@@ -147,11 +147,13 @@ where
 			return false;
 		};
 
-		let candidate = vote_candidate::<B>(number, self.best_grandpa_block, best_beefy_block, self.min_block_delta);
+		let target = vote_target::<B>(self.best_grandpa_block, best_beefy_block, self.min_block_delta);
 
-		metric_set!(self, beefy_should_vote_on, candidate);
+		trace!(target: "beefy", "🥩 should_vote_on: #{:?}, next_block_to_vote_on: #{:?}", number, target);
 
-		number == candidate
+		metric_set!(self, beefy_should_vote_on, target);
+
+		number == target
 	}
 
 	fn sign_commitment(&self, id: &P::Public, commitment: &[u8]) -> Result<P::Signature, error::Crypto<P::Public>> {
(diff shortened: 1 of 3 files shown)


beefy-gadget/src/worker.rs
```diff
@@ -402,3 +397,46 @@ where
...
+
+	// This replicates the `should_vote_on()` core algorithm.
+	// We use that as an interim solution, until we can test `BeefyWorker` proper
+	fn should_vote_on(number: u64, best_grandpa: u64, best_beefy: u64, min_delta: u64) -> bool {
+		let diff = best_grandpa.saturating_sub(best_beefy);
+		let diff = diff.saturated_into::<u32>();
+		let next_power_of_two = (diff / 2).next_power_of_two() as u64;
+		let next_block_to_vote_on = best_beefy + min_delta.max(next_power_of_two);
+		number == next_block_to_vote_on
+	}
```
tomusdrw, line 413: No, duplicating the code is not the way to go.

Please move this function one scope up (simply to the `worker` module) and change `BeefyWorker::should_vote_on` to simply call this function. Rough implementation:
```
impl BeefyWorker {
...
  fn should_vote_on(&self, number: NumberFor<B>) -> bool {
    should_vote_on(number, self.best_grandpa_block, self.best_beefy_block, self.min_block_delta)
  }
...
}
```

tomusdrw (changes requested):

beefy-gadget/src/worker.rs
adoerr: Damn, should have read the book ;-)

beefy-gadget/src/worker.rs
```diff
@@ -402,3 +384,133 @@ where
...
+	use super::vote_candidate;
+	use sp_runtime::testing::{Block, ExtrinsicWrapper, Header};
+
+	type MockBlock = Block<ExtrinsicWrapper<u64>>;
+
+	macro_rules! block {
+		($n:expr) => {
+			Header::new_from_number($n).number;
+		};
+	}
```
tomusdrw, line 427: I'm confused a bit, why is it a macro and not regular function? Also why do we create header and then only take `number` from it?
Wouldn't something like this work:
```rust
fn block(n: u32) -> NumberFor<MockBlock> {
  n.into()
}
```

beefy-gadget/src/worker.rs
```diff
@@ -402,3 +384,133 @@ where
...
+	number: NumberFor<B>,
+	best_grandpa: NumberFor<B>,
+	best_beefy: NumberFor<B>,
+	min_delta: u32,
+) -> NumberFor<B>
+where
+	B: Block,
+{
+	let diff = best_grandpa.saturating_sub(best_beefy);
+	let diff = diff.saturated_into::<u32>();
```
tomusdrw, line 401: Note that the `candidate` must always be lower than `best_grandpa`. That was actually the reason we were doing `/ 2` in the previous code. Currently it will be possible to vote on `candidate` only iff `diff` is a power of two (as seen in the test cases).

beefy-gadget/src/worker.rs
```diff
@@ -402,3 +384,133 @@ where
...
+		assert_eq!(1512, c);
+
+		let c = vote_candidate::<MockBlock>(block!(1024), block!(1024), block!(0), 4);
+		assert_eq!(1024, c);
+	}
+
+	#[test]
+	fn vote_on_target_block() {
+		let c = vote_candidate::<MockBlock>(block!(1008), block!(1008), block!(1002), 4);
+		assert_eq!(1010, c);
```
tomusdrw, line 492: `candidate` blocks should not be greater than `best_grandpa` blocks. We must not vote for non-existent or non-finalized blocks.

tomusdrw (changes requested):
Almost perfect, but we need to revert `diff / 2` back.

beefy-gadget/src/worker.rs
```diff
@@ -142,32 +140,18 @@ where
...
-			next_power_of_two,
-			next_block_to_vote_on,
-		);
+		let candidate = vote_candidate::<B>(number, self.best_grandpa_block, best_beefy_block, self.min_block_delta);
 
-		metric_set!(self, beefy_should_vote_on, next_block_to_vote_on);
+		metric_set!(self, beefy_should_vote_on, candidate);
 
-		number == next_block_to_vote_on
+		number == candidate
```
adoerr, line 154: This line will prevent to vote on a candidate which is ahead of `number`.

beefy-gadget/src/worker.rs
adoerr: We will not. The expr `number == candidate` in the `should_vote_on()` does prevent that. Also running a local test node shows that we vote exactly as we should.

adoerr: The block number in the `assert_eq` is the `best_grandpa` number we would vote for.

beefy-gadget/src/worker.rs
adoerr: If we keep the `/ 2` we will never catch up is the delta between `best_grandpa` and `best_beefy` becomes larger than `min_delta`. Please see the previous commit and the tests.

The reason is that div 2 is basically the same as a minus 1 from the exponent. So we are calculating basically one power of two 'to less' such to speak.

beefy-gadget/src/worker.rs
adoerr: Actually finding this test header thing took me ages so did the first thing that came to mind. And I like macros ;-)

tomusdrw: Hmm, but this code can really be minimized to `$n.into()`, can't it? I'm trying to avoid macros cause they are magical (like now we actually have to discuss what the code is supposed to do), but can live with it as syntactic sugar, however if it can be minimized, let's do it.

andresilva (changes requested):

beefy-gadget/src/worker.rs
tomusdrw: Yes, you're right apologies. I had incorrect understanding how the code works (i.e. what does it mean to "select a candidate") and this is why I was suggesting the changes. More context is available in this issue: https://github.com/paritytech/grandpa-bridge-gadget/issues/162

beefy-gadget/src/worker.rs
tomusdrw: Yeah, renaming might be good. To me it's basically: `next_block_to_vote_on`.

beefy-gadget/src/worker.rs
tomusdrw: Makes sense! Some more context in https://github.com/paritytech/grandpa-bridge-gadget/issues/162

beefy-gadget/src/worker.rs
adoerr: Agreed, the macro really isn't adding any value here.

beefy-gadget/src/worker.rs
```diff
@@ -402,3 +384,133 @@ where
 
 	header.digest().convert_first(|l| l.try_to(id).and_then(filter))
 }
+
+/// Calculate the candidate block for the next BEEY vote.
+///
+/// Note, `number` is only used for better tracing output.
+fn vote_candidate<B>(
+	number: NumberFor<B>,
```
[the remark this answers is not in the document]
adoerr: `number` was used for tracing. Removed it as a parameter and reworked the tracing

beefy-gadget/src/worker.rs
```diff
@@ -402,3 +384,133 @@ where
 
 	header.digest().convert_first(|l| l.try_to(id).and_then(filter))
 }
+
+/// Calculate the candidate block for the next BEEY vote.
+///
+/// Note, `number` is only used for better tracing output.
+fn vote_candidate<B>(
```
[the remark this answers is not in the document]
adoerr: Tried to use more restrictive trait bounds. Couldn't make it work in a way that is actually more succinct.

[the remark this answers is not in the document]
andresilva: Did you try using `BaseArithmetic` or `AtLeast32Bit`? The way it is right now requires that you always annotate the block type when calling the function since it cannot be deduced from the parameters.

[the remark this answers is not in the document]
adoerr: I tried `BaseArtihmetic`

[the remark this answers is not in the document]
andresilva: Pushed a change to address this.

andresilva (approved):

andresilva:
Let's get this into polkadot and then ask Erin to update the wococo nodes. Once this PR is merged `cargo update -p beefy-primitives` in polkadot should be enough.

tomusdrw (approved):

PR #175, opened 2021-05-03
Use resolver v2

adoerr:


tomusdrw (approved):
Shouldn't that regenerate `Cargo.lock`?

adoerr:
I think `Cargo.lock` is not changes, because using `v2` did actually not change anything.

A change would most likely be caused by `[dev-dependencies]`  which are also used as `[dependencies]`. ATM, we don't have that many `[dev-dependencies]`. Those we do have, don't seem to be used as `[dependencies]` also.

Issue #176, opened 2021-05-03
Append SignedCommitment to Block justifications


adoerr:
A BEEFY `SignedCommitment` should be appended to the  justifications for the Block for which the `SignedCommitment` has been generated.

PR #177, opened 2021-05-04
Append SignedCommitment to block justifications

adoerr:
Resolves: #176 
Resolves: #17 

See also: [chain.getBlock justification display byte ordering issue](https://github.com/polkadot-js/apps/issues/5202)

### DIFF
--- a/beefy-gadget/src/lib.rs
+++ b/beefy-gadget/src/lib.rs
@@ -84,6 +84,7 @@ where
 	// empty
 }
 
+/// BEEFY gadget initialization parameters.
 pub struct BeefyParams<B, P, BE, C, N>
 where
 	B: Block,
@@ -147,7 +148,7 @@ where
 			}
 		});
 
-	let worker = worker::BeefyWorker::<_, _, BE, P>::new(
+	let worker_params = worker::WorkerParams {
 		client,
 		backend,
 		key_store,
@@ -156,7 +157,9 @@ where
 		gossip_validator,
 		min_block_delta,
 		metrics,
-	);
+	};
+
+	let worker = worker::BeefyWorker::<_, _, _, _>::new(worker_params);
 
 	worker.run().await
 }
--- a/beefy-gadget/src/lib.rs
+++ b/beefy-gadget/src/lib.rs
@@ -26,7 +26,6 @@ use sc_network_gossip::{GossipEngine, Network as GossipNetwork};
 use sp_api::ProvideRuntimeApi;
 use sp_application_crypto::AppPublic;
 use sp_blockchain::HeaderBackend;
-use sp_consensus::SyncOracle as SyncOracleT;
 use sp_keystore::SyncCryptoStorePtr;
 use sp_runtime::traits::Block;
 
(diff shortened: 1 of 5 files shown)


beefy-gadget/src/lib.rs
```diff
@@ -88,8 +88,10 @@ where
 /// Start the BEEFY gadget.
 ///
 /// This is a thin shim around running and awaiting a BEEFY worker.
+#[allow(clippy::too_many_arguments)]
```
tomusdrw, line 91: Maybe let's introduce a config struct instead?

beefy-gadget/src/worker.rs
```diff
@@ -313,6 +317,14 @@ where
 
 				debug!(target: "beefy", "🥩 Round #{} concluded, committed: {:?}.", round.1, signed_commitment);
 
+				if self
+					.backend
+					.append_justification(BlockId::Number(round.1), (BEEFY_ENGINE_ID, signed_commitment.encode()))
```
tomusdrw, line 322: I think it makes sense to introduce one-byte versioning right from start.

beefy-gadget/src/worker.rs
```diff
@@ -313,6 +317,14 @@ where
...
+					.is_err()
+				{
+					error!(target: "beefy", "🥩 Failed to append justification: {:?}", signed_commitment);
```
tomusdrw, line 325: Since this is happening inside a `handle_vote`, what are the guarantees that it is never appended twice? I.e. how do we control the rounds currently?

tomusdrw (approved):
lgtm!

beefy-gadget/src/lib.rs
```diff
@@ -85,18 +84,33 @@ where
 	// empty
 }
 
+pub struct BeefyParams<B, P, BE, C, N>
```
tomusdrw, line 87:
```suggestion
/// BEEFY worker initialization parameters.
pub struct BeefyParams<B, P, BE, C, N>
```

beefy-gadget/src/worker.rs
```diff
@@ -101,8 +102,10 @@ where
 	/// BEEFY pallet has been deployed on-chain.
 	///
 	/// The BEEFY pallet is needed in order to keep track of the BEEFY authority set.
+	#[allow(clippy::too_many_arguments)]
```
tomusdrw, line 105: Couldn't this thing use `BeefyParams` as well?

adoerr: Yep, I'm not done yet :-)

beefy-gadget/src/worker.rs
adoerr: Trying to append it a second time will lead to a [Duplicate consensus engine ID](https://github.com/paritytech/substrate/blob/3f110196163b5ec03bac5ee188d60bedf3ebd91d/client/db/src/lib.rs#L1815) error.

beefy-gadget/src/worker.rs
adoerr: So this would be something like:
```
let version: u8;
... (BEEFY_ENGINE_ID, (version, signed_commitment).encode())
```
?

beefy-gadget/src/worker.rs
adoerr: I have introduced `WorkerParams` as well.

beefy-gadget/src/worker.rs
tomusdrw: Yeah, or just:
```
#[derive(Encode, Decode)]
pub enum VersionedSignedCommitment {
  V1(SignedCommitment),
}
```

which should be easier to work with long term - the resulting encoding though should be identical in any of the solutions afair though.

PR #186, opened 2021-05-18
Optimize encoded-size of `SignedCommitment`

### LINKED ISSUE #128: Use bitfield in `SignedCommitment` for encoded-size optimisation.
Currently `SignedCommitment` uses `Vec<Option<Signature>>>` for the list of signatures. This vector is expected to contain at least 2/3+1 `Some` values anyway.

It's possible to optimize the storage though, and save 7 bits (1byte for enum variant vs 1 bit) per every signature in SCALE encoded form (minus alignment) by using a different scheme of:
```
signatures_from: BitField,
signatures: Vec<Signature>,
```
Every set bit in `signatures_from` signifies the index of validator that the next signature in `signatures` vec is for.

I think this should be implemented as custom SCALE encoding and we should keep the current `struct` format for backward compatibility (code not over-the-wire).

de-sh:
Fixes #128

- Creates a new, better compressed representation of `SignedCommitment`
- Implement `parity_scale_codec::Encode`
- Implement `parity_scale_codec::Decode`


adoerr:
@de-sh can you pls sign the contributors agreement (CLA) above?

de-sh:
> @de-sh can you pls sign the contributors agreement (CLA) above?

Hey, I think I did 😅

I am working with @tomusdrw to prepare the PR for merger.

adoerr:
Looks like the Bot thinks otherwise

de-sh:
The cla marker within the PR checks seem to be marked green.

de-sh:
@tomusdrw [seems like the hex literal](https://github.com/paritytech/grandpa-bridge-gadget/pull/186/commits/07cacba6244187efbcfbc7d296fda0697691ca77#diff-2527b93169a0aa8d7a4adc826818ad4b2809ef316270f4a4d13f8a2cb106e870R285) for compressed version of `SignedCommitment`, can you check if this is proper. All that's left is some documentation.

beefy-primitives/src/commitment.rs
```diff
@@ -77,7 +78,7 @@ where
 }
 
 /// A commitment with matching GRANDPA validators' signatures.
```
tomusdrw, line 80:
```suggestion
/// A commitment with matching GRANDPA validators' signatures.
///
/// Note that SCALE-encoding of the structure is optimized for size efficiency over the wire,
/// please take a look at custom [`Encode`] and [`Decode`] implementations and [`TemporarySignatures`] struct.
```

beefy-primitives/src/commitment.rs
```diff
@@ -95,6 +96,130 @@ impl<TBlockNumber, TPayload, TSignature> SignedCommitment<TBlockNumber, TPayload
...
 }
 
+/// Type to be used to denote placement of signatures
+type BitField = Vec<u8>;
+/// Compress 8 bit values into a single u8 Byte
+const CONTAINER_BIT_SIZE: usize = 8;
+
+/// Temporary representation used for encoding efficiency.
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode)]
+struct TemporarySignatures<TCommitment, TSignature> {
```
tomusdrw, line 106:
```suggestion
struct SizeOptimizedSignedCommitment<TCommitment, TSignature> {
```
I'd change the name to make a better relation with `SignedCommitment`, but I'm not particularly happy with the proposed name. Opening discussion for some brainstorming. Maybe `PackedSignedCommitment` or `CompressedSignedCommitment`?

beefy-primitives/src/commitment.rs
```diff
@@ -95,6 +96,130 @@ impl<TBlockNumber, TPayload, TSignature> SignedCommitment<TBlockNumber, TPayload
...
+	commitment: TCommitment,
+	signatures_from: BitField,
```
tomusdrw, line 108: Would be cool to add docs what every bit signifies and a more general description on the entire struct to cover how things get's packed.

beefy-primitives/src/commitment.rs
```diff
@@ -95,6 +96,130 @@ impl<TBlockNumber, TPayload, TSignature> SignedCommitment<TBlockNumber, TPayload
...
+impl<'a, TBlockNumber, TPayload, TSignature> From<&'a SignedCommitment<TBlockNumber, TPayload, TSignature>>
+	for TemporarySignatures<&'a Commitment<TBlockNumber, TPayload>, &'a TSignature>
+where
+	TSignature: Encode,
+	TBlockNumber: Encode,
+	TPayload: Encode,
+{
+	/// Convert `SignedCommitment`s into `TemporarySignatures` that are packed better for
+	/// network transport.
+	fn from(signed_commitment: &'a SignedCommitment<TBlockNumber, TPayload, TSignature>) -> Self {
```
tomusdrw, line 122: I know that I suggested `From` implementations initially, but I now things that it might be better of as regular `imlpl` blocks. Also I think the bounds are not really necessary at this point, are they?:


```rust
impl<'a, TBlockNumber, TPayload, TSignature>
  TemporarySignatures<&'a Commitment<TBlockNumber, TPayload>, &'a TSignature>
{
  fn pack(signed_commitment: &'a SignedCommitment<TBlockNumber, TPayload, TSignature>) -> Self {
    ...
  }
}

impl<'a, TBlockNumber, TPayload, TSignature>
  TemporarySignatures<Commitment<TBlockNumber, TPayload>, TSignature>
{
  fn unpack(self) -> SignedCommitment<TBlockNumber, TPayload, TSignature> {
    ...
  }
}
```

beefy-primitives/src/commitment.rs
```diff
@@ -157,7 +282,7 @@ mod tests {
 		assert_eq!(
 			encoded,
 			hex_literal::hex!(
-				"3048656c6c6f20576f726c6421050000000000000000000000000000000000000000000000100000011001020304011005060708"
+				"3048656c6c6f20576f726c64210500000000000000000000000000000000000000000000000403040000000810010203041005060708"
```
tomusdrw, line 285: Hmm, this one is actually longer, but I think it's because we only have 4 signatures.

Do you mind adding a test with 1024 signatures and `(1024 * 2 + 2) / 3` of them being set to `Some`?

tomusdrw (commented):
Great, thanks!

beefy-primitives/src/commitment.rs
de-sh: 1024? woah 😂 
Let me see, might take some time for me to get to this, is that fine?

beefy-primitives/src/commitment.rs
adoerr: `CommitmentCodec`?

de-sh: You are right, something understandable, how about just `CompactSignedCommitment`?

beefy-primitives/src/commitment.rs
de-sh: We could do it the way illustrated in your code, removing the bounds doesn't seem to work though...

beefy-primitives/src/commitment.rs
tomusdrw: Just use a loop :) The signature content doesn't really matter - can be the same for each.

beefy-primitives/src/commitment.rs
tomusdrw: Have you tried removing the bounds only for `pack/unpack` methods? They are definitely needed for `Encode/Decode` implementation though.

beefy-primitives/src/commitment.rs
de-sh: Yeah, can you check this [without compression](https://github.com/paritytech/grandpa-bridge-gadget/commit/0d607aaff7da34120c1e7badaf8b4d987fdec101) and [with compression](https://github.com/paritytech/grandpa-bridge-gadget/pull/186/commits/d5d4b8850d6ae15d0c99065b7b263005e49c772a)

beefy-primitives/src/commitment.rs
de-sh: Yes, the wierd compile errors start popping up, let me try this again when I get the time over the weekend.

de-sh:
Hey, so I did a fresh install of the OS on my laptop, now I receive the error: 
```
❯ cargo test
...
  error[E0463]: can't find crate for `std`
    |
    = note: the `wasm32-unknown-unknown` target may not be installed
```

but when I already have it installed... So that is weird
```
❯ rustup component add rust-src
info: component 'rust-src' is up to date
```
any ideas why this might be happening, it used to work as usual earlier...

beefy-primitives/src/commitment.rs
tomusdrw: `CompactSignedCommitment` sounds perfect to me.

beefy-primitives/src/commitment.rs
tomusdrw: Cool, afaict it's 8966 characters vs 7186 so around 20% optimisation, nice!

beefy-primitives/src/commitment.rs
tomusdrw: Make sure to follow the [Build guide](https://github.com/paritytech/grandpa-bridge-gadget/blob/master/README.md#build) from readme.

beefy-primitives/src/commitment.rs
```diff
@@ -95,6 +99,136 @@ impl<TBlockNumber, TPayload, TSignature> SignedCommitment<TBlockNumber, TPayload
...
+			signatures_compact,
+		} = temporary_signatures;
+		let mut bits: Vec<u8> = vec![];
+		let last_block = signatures_len % CONTAINER_BIT_SIZE as u32;
+		for block in signatures_from {
+			let start_bit = if signatures_len > last_block {
+				0
+			} else {
+				CONTAINER_BIT_SIZE - last_block as usize
+			};
```
tomusdrw, line 201: I don't think this condition is correct, it does not depend on any of the loop items. 

Instead I'd suggest we take all elements from `signature_from` except for the last one which might not be full and only then we process the last item up to the number of bits defined in `last_block`.

beefy-primitives/src/commitment.rs
```diff
@@ -95,6 +99,136 @@ impl<TBlockNumber, TPayload, TSignature> SignedCommitment<TBlockNumber, TPayload
...
+
+			for bit in start_bit..CONTAINER_BIT_SIZE {
+				let bit_position = CONTAINER_BIT_SIZE - bit - 1;
+				bits.push(block >> bit_position & 1);
+				signatures_len -= 1;
+			}
```
tomusdrw, line 207: Actually, disregard my previous comment - given the way we process `bits` later on, I think it's safe to not do any conditionals for the last block - it will simply be filled with `0`.

So the way I see this loop is that we just read everything from `signatures_from`. The inner loop should simply go `for bit in 0..CONTAINER_BIT_SIZE` and then we just trim resulting `signatures` vector to match `signatures_len`. That should simplify the code a lot - wdyt?

beefy-primitives/src/commitment.rs
```diff
@@ -95,6 +99,136 @@ impl<TBlockNumber, TPayload, TSignature> SignedCommitment<TBlockNumber, TPayload
...
+
+impl<TBlockNumber, TPayload, TSignature> Decode for SignedCommitment<TBlockNumber, TPayload, TSignature>
+where
+	TBlockNumber: Decode,
+	TPayload: Decode,
+	TSignature: Decode,
+{
+	fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+		let temp = CompactSignedCommitment::decode(input)?;
+		Ok(CompactSignedCommitment::unpack(temp))
```
tomusdrw, line 228: I just realised that we already have chains that store signed commitments in the database (Rococo & Wococo).
We may either restart them or we could introduce a backward compatible reading here temporarily (until they are restarted).

@andresilva @adoerr wdyt? 

A backward compatible suggestion would look like this (note we should add a test for this):

```suggestion
		let temp = match CompactSignedCommitment::decode(input) {
		  Ok(temp) => Ok(CompactSignedCommitment::unpack(temp)),
		  Err(_) => {
		     type OldSignedCommitment = (Commitment, Vec<Option<Signature>>);
		      let (commitment, signatures) = OldSignedCommitment::decode(input)?;
		      log::warn!("Decoded `SignedCommitment` in old format, this is not supported and will be removed soon. Please restart or migrate your chain.");
		      Ok(SignedCommitment { commitment, signatures })
		  }
		};
```
I guess we might need to do some magic to clone the `input` though, but it should be doable.

tomusdrw (commented):
Looks good, few improvements noted.

beefy-primitives/src/commitment.rs
tomusdrw: @de-sh Please let me know if you need any more clarification on this. I think this is definitely worth doing and will simplify the code a lot.

beefy-primitives/src/commitment.rs
tomusdrw: After discussing with @andresilva and @adoerr we've figured that we can work around backward compatibility issue. So please disregard the comment.

beefy-primitives/src/commitment.rs
de-sh: Awesome! Now I understand quite a lot better, thank you for the patience 😃

tomusdrw: Try this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=dd88ffe4bc09e1d826ea5a6d9c6d0b90

tomusdrw: Instead of doing a slice we can also store the bits in reverse order and truncate (but then we also need to store compact signatures in reverse order too).

de-sh: Just noticed what looks like an edge case: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9c8971e090acacd84518149533e23e04

de-sh: This seems like a potential solution, I'll push it, please review.

beefy-primitives/src/commitment.rs
de-sh: I have tried a few stuff, but the same issues that plagued me earlier with lifetime annotation keep coming to haunt here... Could you take a look? If not, seems to be resolved.

de-sh: This was a big misconception on my side, seems like I got it figured, haha

beefy-primitives/src/commitment.rs
```diff
@@ -95,6 +99,127 @@ impl<TBlockNumber, TPayload, TSignature> SignedCommitment<TBlockNumber, TPayload
...
+			}
+		}
+		let last_byte_len = signatures_len as usize % CONTAINER_BIT_SIZE;
+		let bits = if last_byte_len != 0 {
+			[
+				&bits[..bits.len() - CONTAINER_BIT_SIZE],
+				&bits[bits.len() - last_byte_len..],
+			]
+			.concat()
+		} else {
```
tomusdrw, line 185: We should be avoiding any special casing the last byte, there is definitely a solution that does not involve that.

Basically currently we kind of mess up the bit ordering (numbers are signature indexes):
```
[
  3 2 1 0, // byte 1
  7 6 5 4, /// byte 2
  ...
]
```
This is obviously not wrong, but makes it a bit more difficult to reason about and to propose a proper solution for the last byte issue.

I'd suggest to change the way bits are encoded, so that visually it matches what is in signatures vector. Then after converting to `bits` vector we should make sure the order is correct as well, so that simply truncating works.

de-sh: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9ee8f2399283a346595a205697cd1b04

Did you check this?

tomusdrw: Oh, missed this somehow but it looks really good, I'd go in this direction.

de-sh:
Hey, so I wrote a test to ensure that future contributors don't perform updates that remove constraints regarding off-eight signature(len>8 & len%8 != 0). Do you feel we are ready to merge?

tomusdrw (approved):
lgtm! Have you considered using `bitvec` crate?

de-sh:
Not really, I was trying to achieve a simple implementation without adding external dependencies. Would that have been the right way to go?

tomusdrw:
No, I was asking just out of curiosity. I guess it would simplify bit conversion quite a lot, but since it's already solved I'm happy to avoid the extra dependency.

de-sh:
Tried merging changes from `master` into here, need help with lifetime magic 😅

seunlanlege:
looks like this pr needs to go to substrate

adoerr:
This has not been merged on purpose so far. Merging this will break Snowforks's `SnowBridge`

seunlanlege:
> Merging this will break Snowforks's SnowBridge

They're big boys, they can handle it 😂 @musnit

musnit:
> > Merging this will break Snowforks's SnowBridge
> 
> They're big boys, they can handle it 😂 @musnit

:D yeh once this is finalized and ready we can start working on including it

de-sh:
> looks like this pr needs to go to substrate, you want to handle that @de-sh

Can you lemme know what I can do?

tomusdrw:
> Can you lemme know what I can do?

@de-sh apologies for the delay. Could you please try to apply exactly the same code changes as in this PR, but to the https://github.com/paritytech/substrate/ repository? BEEFY code resides now there. I'm willing to fast-track this change this time to make sure we merge this before any real-world deployment.

de-sh:
Closing in favor of paritytech/substrate#10409

tomusdrw:
Thank you @de-sh!!!

PR #191, opened 2021-05-31
Document BEEFY key

adoerr:


README.md
```diff
@@ -73,6 +73,38 @@ be utilized by a light client implementation.
 │  └──  ...
  ```
 
+## BEEFY Key
+
+The current cryptographic scheme used by BEEFY is `ecdsa`. This is **different** from other keys being used in Substrate. Other
+Substrate related cryptographic keys are using the `sr25519` scheme. The most noticeable difference is that an `ecdsa` public key
```
tomusdrw, line 79:
```suggestion
The current cryptographic scheme used by BEEFY is `ecdsa`. This is **different** from other schemes like `sr25519` and `ed25519` which are commonly used in Substrate configuration for other pallets (BABE, GRANDPA, AuRa, etc). The most noticeable difference is that an `ecdsa` public key 
```

tomusdrw: While the previous is not technically incorrect, it's not like Substrate is bound to `sr25519`. It's just it was chosen over other crypto for default configuration, so it's commonly used, but the crypto is configurable.

README.md
```diff
@@ -73,6 +73,38 @@ be utilized by a light client implementation.
 │  └──  ...
  ```
 
+## BEEFY Key
+
+The current cryptographic scheme used by BEEFY is `ecdsa`. This is **different** from other keys being used in Substrate. Other
+Substrate related cryptographic keys are using the `sr25519` scheme. The most noticeable difference is that an `ecdsa` public key 
+is `33` bytes long, instead of `32` bytes for a `sr25519` based public key. So, a BEEFY key [sticks out](https://github.com/paritytech/polkadot/blob/25951e45b1907853f120c752aaa01631a0b3e783/node/service/src/chain_spec.rs#L738) among the other public keys a bit. The easiest way to generate a BEEFY key is by using the [Subkey](https://substrate.dev/docs/en/knowledgebase/integrate/subkey) tool.
```
tomusdrw, line 80:
```suggestion
is `33` bytes long, instead of `32` bytes for a `sr25519` based public key. So, a BEEFY key [sticks out](https://github.com/paritytech/polkadot/blob/25951e45b1907853f120c752aaa01631a0b3e783/node/service/src/chain_spec.rs#L738) among the other public keys a bit. For other crypto (in default Substrate configuration) the `AccountId` (32-bits) matches the `PublicKey`, but note that it's not the case for BEEFY. The easiest way to generate a BEEFY key is by using the [Subkey](https://substrate.dev/docs/en/knowledgebase/integrate/subkey) tool. 
```

tomusdrw (approved):

Issue #82, opened 2021-02-09
Allow hasher configuration before signing with app-crypto


tomusdrw:
Substrate's `app-crypto` crate allows you to choose from different cryptography and generates application-specific `Public`, `Signature`, etc types.

In BEEFY we currently use `ECDSA` (over `secp256k1`) to make it the most compatible with Ethereum.

However, where we sign the `Commitment` (or rather the encoding of it) using [`SyncCryptoStore::sign_with` here](https://github.com/paritytech/grandpa-bridge-gadget/blob/9f4a049617ccc962c41b90f0b2c8c24c6380c2d4/beefy-gadget/src/lib.rs#L287), it would eventually end up [here](
https://github.com/paritytech/substrate/blob/07a5f2984654a5b8050ec0c0df69b8f081aa75e2/primitives/core/src/ecdsa.rs#L481).

This means that whatever message we are signing (in our case it's the `Commitment`) is going to be unconditionally `blake2_256` hashed before being signed.

Ideally for maximum Ethereum compatibility we would prefer this hash to be `keccak256` instead.

A solution is to either make the hasher somehow configurable in Substrate/`CryptoStore` or fall back to some other mechanism that will allow us to configure how the message is hashed before signing.

tomusdrw:
We can introduce `ecsda_sign_prehashed` to `CryptoStore`, cause it already contain some crypto-specific methods.
@bkchr wdyt?

bkchr:
You mean as host function and as function for the keystore?

Sounds reasonable to me 👍

tomusdrw:
We don't even need a host function, this is purely client code, so it's just about adding to `LocalKeystore` and the trait.

adoerr:
and impl `ecdsa::Pair::sign_prehash()`

PR #192, opened 2021-06-01
Introduce BeefyKeystore

adoerr:
Very much WIP, don't waste time looking at this right now.

`BeefyKeystore` allows to implement crypto scheme specific functionality. 

Resolves: #82

### DIFF
--- a/beefy-gadget/src/gossip.rs
+++ b/beefy-gadget/src/gossip.rs
@@ -105,6 +105,7 @@ where
 		mut data: &[u8],
 	) -> GossipValidationResult<B::Hash> {
 		if let Ok(msg) = VoteMessage::<MmrRootHash, NumberFor<B>, P::Public, P::Signature>::decode(&mut data) {
+			// TODO [ToDr] This should be using BeefyKeystore.
 			if P::verify(&msg.signature, &msg.commitment.encode(), &msg.id) {
 				return GossipValidationResult::ProcessAndKeep(self.topic);
 			} else {
--- a/beefy-gadget/src/keystore.rs
+++ b/beefy-gadget/src/keystore.rs
@@ -52,15 +52,10 @@ impl BeefyKeystore<ecdsa::Pair> for std::sync::Arc<dyn SyncCryptoStore> {
 			.ok_or_else(|| error::Error::Signature("ecdsa_sign_prehashed() failed".to_string()))?;
 
 		// check that `sig` has the expected result type
-		let sig = match sig.clone().try_into() {
-			Ok(s) => s,
-			_ => {
-				return Err(error::Error::Signature(format!(
-					"invalid signature {:?} for key {:?}",
-					sig, public
-				)))
-			}
-		};
+		let sig = sig
+			.clone()
+			.try_into()
+			.map_err(|_| error::Error::Signature(format!("invalid signature {:?} for key {:?}", sig, public)))?;
 
 		Ok(sig)
 	}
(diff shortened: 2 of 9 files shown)


beefy-gadget/src/round.rs
```diff
@@ -53,51 +58,51 @@ fn threshold(authorities: usize) -> usize {
...
+	validator_set: ValidatorSet<P::Public>,
 }
 
-impl<Hash, Number, Id, Signature> Rounds<Hash, Number, Id, Signature>
+impl<H, N, P> Rounds<H, N, P>
 where
-	Hash: Ord,
-	Number: Ord,
+	H: Ord,
+	N: Ord,
```
tomusdrw, line 74: I know it's a never ending discussion and we are not super consistent with this, but I prefer full names for generic parameters instead of one-letter's. It's maybe a bit longer to write, but at least you know roughly what is expected.
I know that the typical counter-argument is that it may be confused with regular types, and I usually suggest something like `THash, TNumber` in such case.

Not going to die on this hill though, but just wanted to leave my 2 cents here.

adoerr: I know, we discussed this earlier ;-) A lot of preferences usually come down to which tooling (esp. editor) one is using. 

But I'm anyways more and more in favour of using associated types instead of generic type parameters wherever possible. And with associated types longer, more descriptive names are unanimous :-)

beefy-gadget/src/gossip.rs
```diff
@@ -105,6 +105,7 @@ where
 		mut data: &[u8],
 	) -> GossipValidationResult<B::Hash> {
 		if let Ok(msg) = VoteMessage::<MmrRootHash, NumberFor<B>, P::Public, P::Signature>::decode(&mut data) {
+			// TODO [ToDr] This should be using BeefyKeystore.
 			if P::verify(&msg.signature, &msg.commitment.encode(), &msg.id) {
```
tomusdrw, line 109: This is not addressed yet.

beefy-gadget/src/keystore.rs
```diff
@@ -0,0 +1,160 @@
...
+pub trait BeefyKeystore<P>: Sync + Send + 'static
+where
+	P: Pair,
+{
+	fn local_id(&self, keys: &[P::Public]) -> Option<P::Public>;
+
+	fn sign(&self, public: &P::Public, message: &[u8]) -> Result<P::Signature, error::Error>;
+
+	fn verify(&self, public: &P::Public, sig: &P::Signature, message: &[u8]) -> bool;
+}
```
tomusdrw, line 33: Please add docs.

beefy-node/node/src/service.rs
```diff
@@ -248,21 +248,21 @@ pub fn new_full(mut config: Configuration) -> Result<TaskManager, ServiceError>
 		None
 	};
 
-	let beefy_params = beefy_gadget::BeefyParams {
+	let beefy_params = beefy_gadget::BeefyParams::<_, beefy_primitives::ecdsa::Pair, _, _, _> {
 		client,
 		backend,
-		key_store: keystore.clone(),
+		// TODO [ToDr]  this is shit we have an Arc<Arc<>> here, I'm sure we can do better.
+		key_store: keystore.as_ref().map(|x| Arc::new(x.clone()) as _),
```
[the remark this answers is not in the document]
tomusdrw: Maybe we can add a helper method at least so that this nightmare is not replicated in every downstream service.
Something like `BeefyParams::cast_keystore(keystore)`

beefy-primitives/src/lib.rs
```diff
@@ -51,6 +51,8 @@ pub mod ecdsa {
 		app_crypto!(ecdsa, crate::KEY_TYPE);
 	}
 
+	pub use self::app_ecdsa::*;
+
```
tomusdrw, line 55: Maybe instead let's not wrap `app_crypto!` invocation into `mod app_ecdsa`.

tomusdrw:
Superseded by #207

Issue #198, opened 2021-06-10
Change commitment payload to make it possible to easily add new items in the future.


tomusdrw:
Currently we sign on MMR root only. In the future we know that there will most likely be MMB (Merkle Mountain Belt) implementation coming, or potentially MMR with different hash usage (non-keccak).
It would be nice if we could keep adding things to the commitment without breaking compatibility with older applications.

We could consider encoding something like:
```
payload: Vec<([u8; 4] /* id */, Vec<u8> /* payload */)>
```
And additionally we should enforce the payload to be sorted by `id` so that it's easy to find the payload one cares about.
Alternatively it can just be a list with fixed ordering (i.e. MMR payload always expected at position `0`).

tomusdrw:
Regarding MMB Alfonso from W3F is the main contact in case w start implementing this :)

seunlanlege:
I think a fixed position saves on unnecessary bytes. should i go ahead and implement this?

tomusdrw:
I'm thinking that the solution with `ids` would be much more flexible and would have less SCALE-compatibility issues long-term. Imagine that we have 5 different payloads which don't necessarily need to be supported by all chains. This means that we always have a `Vec` of 5 elements (older implementations might have a `Vec` of less elements though), and on some entries we will have `None` elements. With `ids` we will have few extra bytes, but the vector stays minimal and we don't have additional `None` entries.

We could maybe use 2-byte id instead though, cause 4-bytes seems like an overkill.

seunlanlege:
closed?

tomusdrw:
Closed via https://github.com/paritytech/substrate/pull/10307

tomusdrw:
CC @musnit @vgeddes

PR #207, opened 2021-06-14
Remove external crypto trait bounds

adoerr:
Alternative implementation for #192.

- Refactor BEEFY crypto by getting rid of crypto related trait bounds.
- Implement `BeefyKeystore` as a newtype

Resolves: #82 

I have deactivated the light client tests for now.

Reading @tomusdrw review comments, I should have provided more context about the reasoning behind this PR right away. Let me do this now. I also changed the PR title to make the intention more clear.

First off, we all agree and it is well understood that the crypto BEEFY uses **internally** should be configurable. This PR is basically a first step in a two step process, refactoring BEEFY crypto in order to provide that configuration option. As such, the PR may seem a step backwards, which it actually is, kind of. 

Starting point were the struggles and the kind of code that had to be written for PR #192. Main root cause for the problems is the fact that BEEFY internal crypto is sort of exposed to the outside by introducing a crypto related trait bound on it's public interface (`sp_crypto::Pair`). IMO, what crypto BEEFY is using should be a purely BEEFY internal property. So, main goal of this PR is to remove the crypto related trait bound on BEEFY's public interface. The idea is, that after this PR is merged, BEEFY is functionally equivalent as to before the PR, by using ECDSA crypto.

But how can we configure BEEFY crypto now? BEEFY crypto is configured through the arguments used for the `app_crypto!()` macro. That is basically configuration through code generation. BEEFY code itself is typed against `beefy_primitives::crypto::{Pair, Public, Signature}`. Those types are newtype wrappers around concrete crypto types generated by the `app_crypto!()` macro.

Why is `BeefyKeystore` not generic? Due to the fact that BEEFY crypto is now configurable through code generation, there are multiple options, how to expose that configuration option. IMO, at this point in time, it is not clear which one is the most suitable. I think that the different options warrant a bit of experimentation and exploration. 

One interesting point is the question, whether a different kind of crypto should result in a new BEEFY protocol, including it's own engine id. Or whether using different gossip topics etc. is enough. Basically, the implications of running multiple BEEFY instances, using different crypto have not been investigated so far.

Sure, we can make `BeefyKeystore` generic right now by adding a type parameter. But so what? It will (or actually should) not be used and might turn out to be we wrong approach, anyway. IMO, it will take a lot more to make configurable BEEFY crypto work seamless.

The signature related unit tests don't use `BeefyKeystore` because it is `beefy_gadget` private. That could be changed and the tests could be using `BeefyKeystore`.

So the tl;dr version of this PR is: do less, in order to enable more.

### DIFF
--- a/beefy-gadget/src/keystore.rs
+++ b/beefy-gadget/src/keystore.rs
@@ -36,7 +36,7 @@ impl BeefyKeystore {
 	///
 	/// Return the public key for which we also do have a private key. If no
 	/// matching private key is found, `None` will be returned.
-	fn authority_id(&self, keys: &[Public]) -> Option<Public> {
+	pub fn authority_id(&self, keys: &[Public]) -> Option<Public> {
 		let store = if let Some(store) = self.0.clone() {
 			store
 		} else {
@@ -57,7 +57,7 @@ impl BeefyKeystore {
 	/// Note that `message` usually will be pre-hashed before being singed.
 	///
 	/// Return the message signature or an error in case of failure.
-	fn sign(&self, public: &Public, message: &[u8]) -> Result<Signature, error::Error> {
+	pub fn sign(&self, public: &Public, message: &[u8]) -> Result<Signature, error::Error> {
 		let store = if let Some(store) = self.0.clone() {
 			store
 		} else {
@@ -83,7 +83,7 @@ impl BeefyKeystore {
 	/// Use the `public` key to verify that `sig` is a valid signature for `message`.
 	///
 	/// Return `true` if the signature is authentic, `false` otherwise.
-	fn verify(public: &Public, sig: &Signature, message: &[u8]) -> bool {
+	pub fn verify(public: &Public, sig: &Signature, message: &[u8]) -> bool {
 		let msg = keccak_256(message);
 		let sig = sig.as_ref();
 		let public = public.as_ref();
--- a/beefy-gadget/src/keystore.rs
+++ b/beefy-gadget/src/keystore.rs
@@ -37,11 +37,7 @@ impl BeefyKeystore {
 	/// Return the public key for which we also do have a private key. If no
 	/// matching private key is found, `None` will be returned.
 	pub fn authority_id(&self, keys: &[Public]) -> Option<Public> {
-		let store = if let Some(store) = self.0.clone() {
-			store
-		} else {
-			return None;
-		};
+		let store = self.0.clone()?;
 
 		for key in keys {
 			if SyncCryptoStore::has_keys(&*store, &[(key.to_raw_vec(), KEY_TYPE)]) {
--- a/beefy-gadget/src/keystore.rs
+++ b/beefy-gadget/src/keystore.rs
@@ -1,0 +1,210 @@
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use std::convert::TryInto;
+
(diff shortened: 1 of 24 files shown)


beefy-primitives/tests/basic_lc_operations.rs
```diff
@@ -14,9 +14,14 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+/*
```
tomusdrw, line 17: uncomment please :)

beefy-primitives/tests/light_client/mod.rs
```diff
@@ -14,10 +14,11 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+/*
```
tomusdrw, line 17: same

beefy-primitives/tests/light_client/validator_set.rs
```diff
@@ -29,7 +29,7 @@ pub enum Signature {
 	Invalid,
 }
 
-impl Signature {
+impl beefy_primitives::crypto::Signature {
```
tomusdrw, line 32: Hmm, okay, I see the issue now. We can't use custom `Public/Signature` key types any more, because we are bound to concrete types.
This is a bit backward change imho and I feel it will give us a hiccup in a near future.

beefy-gadget/src/keystore.rs
```diff
@@ -0,0 +1,206 @@
...
+	crypto::{Public, Signature},
+	KEY_TYPE,
+};
+
+use crate::error;
+
+/// A BEEFY specific keystore implemented as a `Newtype`. This is basically a
+/// wrapper around [`sp_keystore::SyncCryptoStore`] and allows to customize
+/// common cryptographic functionality.
+pub(crate) struct BeefyKeystore(Option<SyncCryptoStorePtr>);
```
tomusdrw, line 30:
```suggestion
pub(crate) struct BeefyKeystore<P: sp_core::Pair /* or sp_application_crypto::AppPair */>(Option<SyncCryptoStorePtr>, PhantomData<P>);
```

beefy-gadget/src/keystore.rs
```diff
@@ -0,0 +1,206 @@
...
+
+impl BeefyKeystore {
```
tomusdrw, line 32:
```suggestion
impl BeefyKeystore<beefy_primitives::crypto::Pair> {
```

It's basically all good, but IMHO we should keep the generic parameter. We know for sure that BEEFY will need to support multiple crypto schemes. We don't have to implement them for now, but people should be free to implement them externally if they like.

tomusdrw (commented):
Nice, but the PR overcomes the issue by **removing** the generic parameter. If we remove the generic parameter then I'm not sure why we even need `BeefyKeystore` abstraction at all, since all the types are know anyway.

IMHO we should keep the current code, but bring back the generic parameter. I know it's a bit of a hassle, but current code will simply break all of the tests that we currently have (and you commented out) - sure we can rewrite them to ecdsa to support current code, but well theses tests are there not really to test if ecdsa signatures are generated and verified correctly, but just  show how light clients are supposed to work. If we add concrete crypto there we will kind of pollute the code with unnecessary extra code imho.

Given that we plan supporting alternative crypto for BEEFY and that's the general vibe for Substrate - before the change people could take BEEFY and configure it with sr25519, it just happens that our specific case needs `ecdsa`. I was imagining that in the future we will simply run beefy worker multiple times and will have multiple crypto being handled independently. Surely we have some more fancy crypto in mind (like aggregated signatures) for the future, that doesn't necessarily fit the public-private-pair type model, but it will require adding a generic parameter anyway.

Having said all that, I'm not 100% that we should really have the generic parameter. It's just my gut feeling that it will make it easier in the future to introduce alternative crypto (by creating a bit of an extra hassle at present), but maybe I'm kind of cargo culting Substrate's approach. 

CC @andresilva what's your thoughts on this?

tomusdrw (approved):
Fair enough. Let's merge and move on, we can revisit later on. We should however draw a clear distinction between BEEFY protocol and BEEFY gadget. While the protocol should be supporting multiple cryptos and we should future proof all of the structures (similar to #198) it doesn't necessarily need to be handled by the same gadget instance (i.e. I was imaging multiple gadgets running independently configured for different crypto requirements).

That said, I'm fine with the current change now and prefer to start addressing things in preparation for the audit.

PR #208, opened 2021-06-15
Add open voting rounds metric

adoerr:
This metric should help to identify issues with not concluding any/enough voting rounds.


adoerr:
Closed, because the metric is actually not really useful, since the number of live rounds is bounded.

PR #209, opened 2021-06-15
Immediate authority change

adoerr:
According to the `SessionHandler::on_new_session()` documentation `changed is true whenever any of the session keys or underlying economic identities or weightings behind those keys has changed.`

We enact such a change now immediately, 

I changed this PR to a draft, because the issue it is supposed to address might have an unrelated root cause. So, this PR might actually not be needed.


tomusdrw:
Closing since I think it's not relevant. Sorry for introducing noise and extra work :)

PR #210, opened 2021-06-17
Add trace points; Reduce MAX_LIVE_GOSSIP_ROUNDS

adoerr:


tomusdrw (approved):
LGTM, can you also add traces that would print out local BEEFY pub keys so that we can compare that with the on-chain validator set?

adoerr:
[This](https://github.com/paritytech/grandpa-bridge-gadget/blob/104e567759cad8f44339d39413eb75e45d77eb5c/beefy-gadget/src/worker.rs#L175) will log the whole new active validator set.

Added [this](https://github.com/paritytech/grandpa-bridge-gadget/blob/0a4f8c1e53e3182cc978fc7a1497210eeeedffb8/beefy-gadget/src/worker.rs#L217) to specifically log the authority id being used for signing the commitment. It is a bit redundant but since we are on `trace` level anyway, that shouldn't be an issue.

tomusdrw:
I actually meant to trace the entire local keystore - for instance to see if there are any BEEFY keys , but not necessarily the ones that match the current authority set.

PR #217, opened 2021-06-21
Additional initial authority id's

adoerr:
- Add `Bob` and `Charlie` as initial authority id's to the chain spec for `beefy-node`. This allows for a much more useful local test net based on `beefy-node`. The instructions for how to run such a test net have been updated as well.

- Improve some trace points

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,10 +109,38 @@ In case your BEEFY keys are using the wrong cryptographic scheme, you will see a
 
 ## Running BEEFY
 
-Currently the easiest way to see BEEFY in action is to run a single dev node like so:
+Currently the easiest way to run BEEFY is to use a 3-node local testnet using `beefy-node`. We will call those nodes `Alice`, `Bob` and
+`Charlie`. Each node will use the built-in development account with the same name, i.e. node `Alice` will use the `Alice` development
+account and so on. Each of the three accounts has been configured as an initial authority at genesis. So, we are using three validators
+for our testnet.
+
+`Alice` is our bootnode is is started like so:
 
 ```
-$ RUST_LOG=beefy=trace ./target/debug/beefy-node --tmp --dev --alice --validator
+$ RUST_LOG=beefy=trace ./target/debug/beefy-node --tmp --alice -d /tmp/alice
 ```
 
-Expect additional (more useful) deployment options to be added soon.
+`Bob` is started like so:
+
+```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -109,10 +109,38 @@ In case your BEEFY keys are using the wrong cryptographic scheme, you will see a
...
+RUST_LOG=beefy=trace ./target/debug/beefy-node --bob --port 30334 -d/ tmp/bob --bootnodes '/ip4/127.0.0.1/tcp/30333/p2p/ALICE_BOOTNODE_ID_HERE'
+```
+
+`Charlie` is started like so:
+
+```
+RUST_LOG=beefy=trace ./target/debug/beefy-node --charlie --port 30334 -d /tmp/charlie --bootnodes '/ip4/127.0.0.1/tcp/30333/p2p/ALICE_BOOTNODE_ID_HERE'
+```
+
+Note that all nodes are using a persistent DB located under `/tmp/xxx`. For `Bob` and `Charlie` this is not strictly necessary, however. The reason for this is the `ALICE_BOOTNODE_ID_HERE` placeholder.
```
tomusdrw, line 135: Is passing bootnodes really necessary if it's on the same machine? These node should discover each other easily, no?

Have you tried running just with `--tmp` and without `--bootnodes`?

beefy-gadget/src/gossip.rs
```diff
@@ -72,6 +72,9 @@ where
 		}
 	}
 
+	/// Note a live voting round
+	///
+	/// This should be called, if we cast a vote for `round`.
```
tomusdrw, line 77: It seems we actually call it upon receiving a vote as well (`handle_vote`).

adoerr: I actually meant that, because `handle_vote` is the only place `note_vote` is called. Will try to formulate the comment more generic.

beefy-gadget/src/gossip.rs
```diff
@@ -72,6 +72,9 @@ where
 		}
 	}
 
+	/// Note a live voting round
+	///
+	/// This should be called, in order to keep tabs on `round`.
```
tomusdrw, line 77: What do you mean by "keep tabs"?

adoerr: `to keep tabs on sth` means to observe / control / keep sth under observation.

But this comment is going to change with the upcoming gossip fix anyways.

tomusdrw (approved):

PR #219, opened 2021-06-22
Fix note_round()

adoerr:
Currently `note_round()` assumes that noted round numbers do increase monotonically. As long as only validators note rounds, this assumption holds.

However, since we also have non-validators noting rounds (through gossiping), the assumption is wrong. As an implication of this, we basically note more or less random round numbers. Noting unordered round numbers, however, basically shuffles the vector of current live rounds. This in turn leads to re-gossiping a large number of messages like forever. Although, never more than `MAX_LIVE_GOSSIP_ROUNDS` of them.

This patch fixes the excess gossip by keeping the `MAX_LIVE_GOSSIP_ROUNDS` most **recent** round numbers live. This also prevents gossip from stale nodes generated during syncing from spreading among peer nodes.

### DIFF
--- a/beefy-gadget/src/gossip.rs
+++ b/beefy-gadget/src/gossip.rs
@@ -149,7 +149,7 @@ where
 
 #[cfg(test)]
 mod tests {
-	use super::GossipValidator;
+	use super::{GossipValidator, MAX_LIVE_GOSSIP_ROUNDS};
 	use sc_network_test::Block;
 
 	#[test]
@@ -169,7 +169,7 @@ mod tests {
 
 		let live = gv.live_rounds.read();
 
-		assert_eq!(live.len(), 3);
+		assert_eq!(live.len(), MAX_LIVE_GOSSIP_ROUNDS);
 
 		assert!(!GossipValidator::<Block>::is_live(&live, 1u64));
 		assert!(GossipValidator::<Block>::is_live(&live, 3u64));
@@ -188,7 +188,7 @@ mod tests {
 
 		let live = gv.live_rounds.read();
 
-		assert_eq!(live.len(), 3);
+		assert_eq!(live.len(), MAX_LIVE_GOSSIP_ROUNDS);
 
 		assert!(GossipValidator::<Block>::is_live(&live, 3u64));
 		assert!(!GossipValidator::<Block>::is_live(&live, 1u64));
@@ -202,7 +202,7 @@ mod tests {
 
 		let live = gv.live_rounds.read();
 
-		assert_eq!(live.len(), 3);
+		assert_eq!(live.len(), MAX_LIVE_GOSSIP_ROUNDS);
 
 		assert!(GossipValidator::<Block>::is_live(&live, 15u64));
 		assert!(GossipValidator::<Block>::is_live(&live, 20u64));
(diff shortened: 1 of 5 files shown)


beefy-gadget/src/gossip.rs
```diff
@@ -143,11 +146,100 @@ where
...
+
+		assert_eq!(live.len(), MAX_LIVE_GOSSIP_ROUNDS);
+
+		assert!(GossipValidator::<Block>::is_live(&live, 15u64));
+		assert!(GossipValidator::<Block>::is_live(&live, 20u64));
+		assert!(GossipValidator::<Block>::is_live(&live, 23u64));
+	}
+
+	#[test]
+	fn note_same_round_twice() {
```
adoerr, line 221: This is another edge case. How do we want to deal with it? We can adapt the check in `note_round()` for this case as well or just ignore it.

It only happens if we are at `MAX_LIVE_GOSSIP_ROUNDS`  and for the  **first** time a round is noted for the second time which is **not** the currently lowest numbered round.

Easiest way to fix this is to just introduce the `off-by-one` bug again and accept that in some cased, we actually keep `MAX_LIVE_GOSSIP_ROUNDS + 1 ` rounds live.

beefy-gadget/src/gossip.rs
```diff
@@ -53,41 +49,48 @@ where
...
 		}
 
-		if let Some(idx) = live_rounds.binary_search(&round).err() {
-			live_rounds.insert(idx, round);
+		// we want to keep `MAX_LIVE_GOSSIP_ROUNDS` *after* noting `round`
+		while live.len() > (MAX_LIVE_GOSSIP_ROUNDS - 1) {
+			let _ = live.remove(0);
+		}
+
+		if let Some(idx) = live.binary_search(&round).err() {
```
tomusdrw, line 92:
```suggestion
		if let Some(idx) = live.binary_search(&round).err() {
			live.insert(idx, round);
		}
		if|while live.len() > MAX_LIVE_GOSSIP_ROUNDS {
			let _ = live.remove(0);
		}
```

I think this is something that is going to be more future proof and is just simpler. In the previous code you still have a case where we remove a round, but then we don't really add one (because it's already there).

beefy-gadget/src/gossip.rs
```diff
@@ -53,41 +49,47 @@ where
...
+		let mut live = self.live_rounds.write();
 
-		// NOTE: ideally we'd use a VecDeque here, but currently binary search is only available on
-		// nightly for `VecDeque`.
-		while live_rounds.len() > MAX_LIVE_GOSSIP_ROUNDS {
-			let _ = live_rounds.remove(0);
+		// we want to keep `MAX_LIVE_GOSSIP_ROUNDS` most recent rounds.
+		if (live.len() == MAX_LIVE_GOSSIP_ROUNDS) && (live[0] > round) {
+			return;
 		}
```
tomusdrw, line 85: I feel this condition is superfluous now.

adoerr: I could now start to argue that the early return is a bit of an optimization. But that is BS, given the number of live rounds we are going to keep around. So, guilty as charged. I just missed it :angry:

tomusdrw: Yeah, I was thinking the same initially, but then suggested this simplified code and I think less code will be much more important than an extra binary search on 5-element collection :)

tomusdrw (approved):

PR #224, opened 2021-06-28
Use LocalKeystore for tests

adoerr:
Turns out that `sc_keystore::LocalKeystore` and `sp_keystore::testing::KeyStore` are quite different from an implementation point of view. So, using the test version of the key store is basically pointless.

Hence, all `BeefyKeystore` test are now using `LocalKeystore`, i.e. the "real" key store.

Also introduced a `public_keys()` function to `BeefyKeystore` which will be used in an upcoming PR for key store sanity checking.

### DIFF
--- a/beefy-gadget/src/keystore.rs
+++ b/beefy-gadget/src/keystore.rs
@@ -11,7 +11,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-use std::convert::TryInto;
+use std::convert::{From, TryInto};
 
 use sp_application_crypto::RuntimeAppPublic;
 use sp_core::keccak_256;
@@ -23,6 +23,19 @@ use beefy_primitives::{
 };
 
 use crate::error;
+
+// Keystore newtype helper
+//
+// Return wrapped keystore instance or an error in case there is no keystore.
+macro_rules! keystore {
+	($self:expr) => {
+		if let Some(store) = $self.0.clone() {
+			store
+		} else {
+			return Err(error::Error::Keystore("no Keystore".to_string()));
+		}
+	};
+}
 
 /// A BEEFY specific keystore implemented as a `Newtype`. This is basically a
 /// wrapper around [`sp_keystore::SyncCryptoStore`] and allows to customize
@@ -74,6 +87,19 @@ impl BeefyKeystore {
 			.map_err(|_| error::Error::Signature(format!("invalid signature {:?} for key {:?}", sig, public)))?;
 
 		Ok(sig)
+	}
+
+	/// Returns a vector of [`beefy_primitives::crypto::Public`] keys which are currently supported (i.e. found
+	/// in the keystore).
+	pub fn public_keys(&self) -> Result<Vec<Public>, error::Error> {
+		let store = keystore!(self);
+
+		let pk: Vec<Public> = SyncCryptoStore::ecdsa_public_keys(&*store, KEY_TYPE)
+			.iter()
+			.map(|k| Public::from(k.clone()))
+			.collect();
+
+		Ok(pk)
 	}
 
 	/// Use the `public` key to verify that `sig` is a valid signature for `message`.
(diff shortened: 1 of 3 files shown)


beefy-gadget/src/keystore.rs
```diff
@@ -24,6 +24,19 @@ use beefy_primitives::{
 
 use crate::error;
 
+// Keystore newtype helper
+//
+// Return wrapped keystore instance or an error in case there is no keystore.
+macro_rules! keystore {
```
tomusdrw, line 30: Why a macro and not a function?

beefy-gadget/src/keystore.rs
```diff
@@ -76,6 +85,20 @@ impl BeefyKeystore {
 		Ok(sig)
 	}
 
+	#[allow(dead_code)]
```
tomusdrw, line 88: I guess it's dead code because it's going to be used in future PR? I'd rather avoid doing anything like that - why can't we add this in the future PR itself? It's really tiny and I fear that if the future PR doesn't end up being created we will be left with this forever.

tomusdrw (changes requested):

beefy-gadget/src/keystore.rs
adoerr: A function wouldn't add any benefit, since the error handling would still be explicit. The idea was to hide the somewhat verbose error handling. That said, the idiomatic version outlined by you should be good enough here.

beefy-gadget/src/keystore.rs
```diff
@@ -24,6 +24,19 @@ use beefy_primitives::{
...
+// Keystore newtype helper
+//
+// Return wrapped keystore instance or an error in case there is no keystore.
+macro_rules! keystore {
+	($self:expr) => {
+		if let Some(store) = $self.0.clone() {
+			store
+		} else {
+			return Err(error::Error::Keystore("no Keystore".to_string()));
+		}
```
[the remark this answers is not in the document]
adoerr: See above comment.

beefy-gadget/src/keystore.rs
adoerr: I agree. There is a bit of a story behind this PR, however. If you look at the sequence of the commits, the PR was intended to implement and actually apply `public_keys()`. While writing a unit test for `public_keys()` it turned out that the testing key store is different from `LocalKeystore` to a point of being detrimental.

So, this PR changed it's nature and became a PR to get rid of the testing key store and use `LocalKeystore` instead. I really did not wanted to use `public_keys()` without knowing it works. On the other hand, I didn't wanted to bloat the PR any more by actually using `public_keys()`, either.

Guess, I'm asking for a waiver here on the promise that the "upcoming" PR will be the **next** PR.

beefy-gadget/src/keystore.rs
```diff
@@ -96,111 +119,171 @@ impl From<Option<SyncCryptoStorePtr>> for BeefyKeystore {
...
+	#[test]
+	fn public_keys_works() {
+		const TEST_TYPE: sp_application_crypto::KeyTypeId = sp_application_crypto::KeyTypeId(*b"test");
+
+		let store = keystore();
+
+		// test keys
+		let _ = SyncCryptoStore::ecdsa_generate_new(&*store, TEST_TYPE, Some(&Keyring::Alice.to_seed()))
+			.ok()
+			.unwrap();
```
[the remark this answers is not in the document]
adoerr: I like the closure here :+1:

beefy-gadget/src/keystore.rs
```diff
@@ -54,11 +54,10 @@ impl BeefyKeystore {
...
 	/// Return the message signature or an error in case of failure.
 	pub fn sign(&self, public: &Public, message: &[u8]) -> Result<Signature, error::Error> {
-		let store = if let Some(store) = self.0.clone() {
-			store
-		} else {
-			return Err(error::Error::Keystore("no Keystore".to_string()));
-		};
+		let store = self
+			.0
+			.clone()
```
tomusdrw, line 59: I guess instead of cloning we could do `.as_ref()` ? We don't really need ownership of that `Arc`, do we?

tomusdrw (approved):

beefy-gadget/src/keystore.rs
adoerr: Using `.as_ref()` ultimately (after the `ok_or_else`) results in a `&Arc<dyn SyncCryptoStore>` instead of `Arc<dyn SyncCryptoStore>` which requires us to use the store as `&**store`. That looks really odd. 

Since `clone()` on an `Arc` is pretty cheap and idiomatic, I think that is clearer.  Although it may technically not really be needed.

Issue #221, opened 2021-06-25
ParachainHeads & BEEFY validators in MMR are incorrect


tomusdrw:
Currently we make a **trie** (modifier base-16 patricia-merkle tree) out of parachain heads and BEEFY validators uncompressed public keys.

It is quite redundant though and not that easy to verify in Solidity.

Instead we should make a regular binary tree out of:
1. `(para_id, para_head)` (to make it easy to figure out which parachain is which)
2. `ethereum addresses` of BEEFY validators (afair: `address = keccak256(uncompressed_public_key)[0..20]`; to make it faster to verify on ETH)

Note that there are no host functions to compute the binary tree root, so we should either consider adding them (cumbersome!) or check if in-runtime implementation is feasible (I'd start from the latter and only optimize as a host function after).

tomusdrw:
Implemented already.

AurevoirXavier:
How about using sparse-merkle-tree?
It will be much easier to verify multiple leaves. And the tree won't be too large in this case.

tomusdrw:
Can you elaborate? What would be the domain of leaves? What advantage you are striving for?

In case of parachains we could potentially use the id, but the tree may still end up being quite large, because afair the parachain ids can be picked arbitrarily. In case of BEEFY validators we are speaking of 2^160 possible leaves, we would need to bring this down obviously, but I don't see much advantage over just regular merkle tree, especially given that a sparse one (if we put addresses into some buckets) becomes attackable, since validators can pick addresses arbitrarily.

hujw77:
If we have 5 BEEFY validators.
For regular binary tree.
```
                      1
          2                       3'
    4'          5                 3
  6    7     8'    9*             3
  A    B     C     D              E
```
For leaf D, 
witness:  `8, 4, 3`
prove: `H(H(4, H(8,9)), 3) == 1`  
This is an efficient way for one leaf inclusion proof.

But if we extend the leaf to be a sparse merkle tree, 
```
                      1
          2                       3
    4'          5           6           7'
  8    9     10*  11*    12*  13'    14   15
  A    B     C     D     E    Ø      Ø    Ø
```
It will be very easy for a multi-leave proof.
For leaves C, D and E
witness:  `4, 13, 7` 
prove: `H(H(4, H(10,11)), H(H(12,13), 7)) == 1`  

For regular binary tree, it also support multi-leaves too, but the indices is not friendly for verify in solidity.
[...]

PR #225, opened 2021-06-28
Use Binary Merkle Tree instead of a trie

tomusdrw:
Related: #221 

Also convert uncompressed BEEFY keys to ETH addresses first to reduce gas costs.


tomusdrw:
CC @philipstanislaus @musnit I think this is now fully matching the JS library. I've added a test case from your test fixture as well.

I'll proceed with changing the implementation in Polkadot now.

philipstanislaus (approved):
LGTM, thanks for implementing this! 🎉

beefy-merkle-root/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
+description = "A no-std/Substrate compatible library to construct binary merkle tree."
+
+[dependencies]
+hex = { version = "0.4", optional = true }
+log = { version = "0.4", optional = true, default-features = false }
+tiny-keccak = { version = "2.0.1", features = ["keccak"], optional = true }
```
adoerr, line 12: Why don't we use version `2.0.2` (the latest) here?

beefy-merkle-root/src/lib.rs
```diff
@@ -0,0 +1,807 @@
...
+	/// This vec contains all inner node hashes necessary to reconstruct the root hash given the
+	/// leaf hash.
+	pub proof: Vec<Hash>,
+	/// Number of leaves in the original tree.
+	///
+	/// This is needed to detect a case where we have an odd number of leaves that "get promoted"
+	/// to upper layers.
+	pub number_of_leaves: usize,
+	/// Index of the leaf the proof is for.
+	pub leaf_index: usize,
```
adoerr, line 136: Are those indices `0` or `1` based?

beefy-merkle-root/src/lib.rs
```diff
@@ -0,0 +1,807 @@
...
+	let mut leaf = None;
+	let iter = leaves.into_iter().enumerate().map(|(idx, l)| {
+		let hash = H::hash(l.as_ref());
+		if idx == leaf_index {
+			leaf = Some(l);
+		}
+		hash
+	});
+
+	struct ProofCollection {
```
adoerr, line 190: I assume this is the path for a single proof, i.e. all the hashes from leaf `position` to the root. This is not a collection of multiple proofs?

beefy-merkle-root/src/lib.rs
```diff
@@ -0,0 +1,807 @@
...
+/// Supported hashing output size.
+///
+/// The size is restricted to 32 bytes to allow for a more optimised implementation.
+pub type Hash = [u8; 32];
+
+/// Generic hasher trait.
+///
+/// Implement the function to support custom way of hashing data.
+/// The implementation must return a [Hash] type, so only 32-byte output hashes are supported.
+pub trait Hasher {
```
adoerr, line 45: Since this trait is public, would it make sense to support conversion to / form, e.g. `sp_core::Hasher`?

adoerr (approved):
Looks good, some minor remarks.

beefy-merkle-root/src/lib.rs
tomusdrw: 0-based, added a comment.

beefy-merkle-root/src/lib.rs
tomusdrw: Yes, we only support single-leaf proof here.

beefy-merkle-root/src/lib.rs
tomusdrw: I'd rather keep the crate not depending on any substrate stuff.

Cargo.toml
```diff
@@ -8,6 +8,7 @@ members = [
 	"beefy-cli",
 	"beefy-gadget",
 	"beefy-gadget/rpc",
+	"beefy-merkle-root",
```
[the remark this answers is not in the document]
tomusdrw: I started with `merkle-tree`, but then the main function was only `merkle_root` so I renamed. Later on added proof verification as well, so I think it kind of makes sense to call it `merkle-tree` now :)

adoerr (approved):

Issue #212, opened 2021-06-17
GossipValidator should limit number of periodic rebroadcast gossip messages


adoerr:
In `GossipValidator::message_allowed()` introduce a limiter for `MessageIntent::PeriodicRebroadcast` gossip messages. 

This change is intended to reduce the frequency at which vote messages are rebroadcast which in turn will result in reduced BEEFY gossip network traffic.

[This](https://github.com/paritytech/substrate/blob/34d4bb0ed47f7ec5ae26add9f9f8f244428a3e02/client/finality-grandpa/src/communication/gossip.rs#L1582-L1585) would be a viable implementation approach.

PR #227, opened 2021-07-02
Cache known votes in gossip

tomusdrw:
This PR introduces a cache for known votes (ideally I think we should be avoiding the `RwLock` stuff). Each time we receive a valid vote for a particular round, we keep it's hash in gossip to quickly approve such votes in the future if they are coming from other peers. AFAICT the `sc_network_gossip` does de-duplicate incoming messages, but only if they are coming from the same peer, not necessarily if they are sent by multiple peers.

My guess is that receiving & verifying the same votes from multiple peers might be causing the high CPU usage we are observing on Rococo, so this cache should alleviate the issue before we implement #212 and #182. 

Also in `validate` function we quickly reject messages that are for old rounds. Previously we were always verifying and accepting such messages, but later on they were considered "expired" by the gossip subsystem, hence removed.

### DIFF
--- a/beefy-gadget/src/gossip.rs
+++ b/beefy-gadget/src/gossip.rs
@@ -125,7 +125,7 @@ where
 		mut data: &[u8],
 	) -> ValidationResult<B::Hash> {
 		if let Ok(msg) = VoteMessage::<MmrRootHash, NumberFor<B>, Public, Signature>::decode(&mut data) {
-			let msg_hash = twox_64(&data);
+			let msg_hash = twox_64(data);
 			let round = msg.commitment.block_number;
 
 			// Verify general usefulness of the message.
--- a/beefy-gadget/src/gossip.rs
+++ b/beefy-gadget/src/gossip.rs
@@ -86,6 +86,7 @@ where
 
 		let mut live = self.known_votes.write();
 
+		#[allow(clippy::map_entry)]
 		if !live.contains_key(&round) {
 			live.insert(round, Default::default());
 		}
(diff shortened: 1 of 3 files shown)


beefy-gadget/src/gossip.rs
```diff
@@ -116,33 +157,35 @@ where
...
 
 			expired
 		})
 	}
 
 	#[allow(clippy::type_complexity)]
 	fn message_allowed<'a>(&'a self) -> Box<dyn FnMut(&PeerId, MessageIntent, &B::Hash, &[u8]) -> bool + 'a> {
-		let live_rounds = self.live_rounds.read();
+		let known_votes = self.known_votes.read();
 		Box::new(move |_who, _intent, _topic, mut data| {
```
adoerr, line 179: According to how [propage()](https://github.com/paritytech/substrate/blob/cf4e320398cf3b8ef8a2c240d7495c25b3b73b08/client/network-gossip/src/state_machine.rs#L131-L138) handles `MessageIntent::PeriodicRebroadcast`, I'm inclined to just drop rebroadcasts. Something like

```rust
		Box::new(move |_who, intent, _topic, mut data| {
			// We drop rebroadcasts. If a peer doesn't know a message, the gossip state machine will
			// treat a rebroadcast as a broadcast. In other words, if the do see a rebroadcast, it is
			// *alwayas* an already known message.
			if intent == MessageIntent::PeriodicRebroadcast {
				return false;
			}

```
I tested this with `rococo-local` and it worked well. Theoretically this should give us a pretty minimal amount of gossip messages.

adoerr (approved):
Looks reasonable for a stop gap solution.

beefy-gadget/src/gossip.rs
tomusdrw: Hmm, but this doesn't guarantee that the round is concluded for everyone really. Until we have nodes requesting missing BEEFY justifications (I guess part of import queue stuff) I think we should be on the safe side of gossiping more, but without risking some sessions not being concluded.

I guess it is a potential workaround for now as well, but given this cache receiving re-broadcasted votes should be cheap.

adoerr: Totally agree, regarding this being an interim solution. Right now, our problem is, we Gossip *way* to much. I think we should give it a try and see how it goes. Right now, we don't produce any BEEFY justifications on Rococo. Guess, it won't be worse than that.

PR #232, opened 2021-07-06
Some key store sanity checks

adoerr:
- added some key store related sanity checks
- reworked logging, so that log level `debug` would allow to follow the protocol

### DIFF
--- a/beefy-gadget/src/gossip.rs
+++ b/beefy-gadget/src/gossip.rs
@@ -15,7 +15,7 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 use codec::{Decode, Encode};
-use log::{debug, trace};
+use log::{debug, trace, warn};
 use parking_lot::RwLock;
 use std::collections::BTreeMap;
 
@@ -82,7 +82,7 @@ where
 	/// We retain the [`MAX_LIVE_GOSSIP_ROUNDS`] most **recent** voting rounds as live.
 	/// As long as a voting round is live, it will be gossiped to peer nodes.
 	pub(crate) fn note_round(&self, round: NumberFor<B>) {
-		trace!(target: "beefy", "🥩 About to note round #{}", round);
+		debug!(target: "beefy", "🥩 About to note round #{}", round);
 
 		let mut live = self.known_votes.write();
 
@@ -149,7 +149,7 @@ where
 				return ValidationResult::ProcessAndKeep(self.topic);
 			} else {
 				// TODO: report peer
-				debug!(target: "beefy", "🥩 Bad signature on message: {:?}, from: {:?}", msg, sender);
+				warn!(target: "beefy", "🥩 Bad signature on message: {:?}, from: {:?}", msg, sender);
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


beefy-gadget/src/keystore.rs
```diff
@@ -39,10 +41,18 @@ impl BeefyKeystore {
...
+			.iter()
+			.filter(|k| SyncCryptoStore::has_keys(&*store, &[(k.to_raw_vec(), KEY_TYPE)]))
+			.cloned()
+			.collect();
+
+		if public.len() > 1 {
+			warn!(target: "beefy", "🥩 Multiple private keys found for: {:?}", public);
+			return Some(public[0].clone());
+		} else if !public.is_empty() {
+			return Some(public[0].clone());
```
adoerr, line 55: I have opted for this check instead of checking the signed commitment right after signing it.

Main reason is that I do see a potential issue with multiple private keys if one is an ephemeral (in-memory) and the other is a persistent private key. The in-memory private key will *always* be the first one returned by *has_keys()*. So it could happen that the sign the commitment with the in-memory private key which will be gone after a node re-start. 

The above check together with `verify_validator_set()` will hopefully allow us to identify those cases.

beefy-gadget/src/worker.rs
```diff
@@ -214,10 +239,10 @@ where
 
 		if self.should_vote_on(*notification.header.number()) {
 			let authority_id = if let Some(id) = self.key_store.authority_id(self.rounds.validators().as_slice()) {
-				trace!(target: "beefy", "🥩 Local authority id: {:?}", id);
+				debug!(target: "beefy", "🥩 Local authority id: {:?}", id);
 				id
 			} else {
-				trace!(target: "beefy", "🥩 Missing validator id - can't vote for: {:?}", notification.header.hash());
+				warn!(target: "beefy", "🥩 Missing validator id - can't vote for: {:?}", notification.header.hash());
```
tomusdrw: We need to tone this down, it's going to be triggered on every non-validator node and there is nothing the user can do to fix this.

beefy-gadget/src/keystore.rs
tomusdrw: It still doesn't cover a case where keystore thinks has the keys, but the keys are simply invalid. I've added a check when `trace` logging is enabled.

beefy-gadget/src/gossip.rs
```diff
@@ -149,7 +149,7 @@ where
 				return ValidationResult::ProcessAndKeep(self.topic);
 			} else {
 				// TODO: report peer
-				debug!(target: "beefy", "🥩 Bad signature on message: {:?}, from: {:?}", msg, sender);
+				warn!(target: "beefy", "🥩 Bad signature on message: {:?}, from: {:?}", msg, sender);
```
tomusdrw: This is not something that depends on local user configuration, we have a convention to not trigger `warn`s for things that are not addressable by the user.

tomusdrw (approved):
I've pushed few changes to tone down the warnings, but looks good otherwise. Let's deploy ASAP.

PR #243, opened 2021-07-19
gadget: add global timeout for rebroadcasting messages

### LINKED ISSUE #212: GossipValidator should limit number of periodic rebroadcast gossip messages

andresilva:
Fix #212.

### DIFF
--- a/beefy-gadget/Cargo.toml
+++ b/beefy-gadget/Cargo.toml
@@ -12,6 +12,7 @@ hex = "0.4"
 log = "0.4"
 parking_lot = "0.11"
 thiserror = "1.0"
+wasm-timer = "0.2.5"
 
 codec = { version = "2.0.0", package = "parity-scale-codec", features = ["derive"] }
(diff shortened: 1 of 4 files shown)


beefy-gadget/Cargo.toml
```diff
@@ -12,6 +12,7 @@ hex = "0.4"
 log = "0.4"
 parking_lot = "0.11"
 thiserror = "1.0"
+wasm-timer = "0.2.5"
```
andresilva, line 15: `std::time::Instant` does not work on wasm target, we need this for the browser node.

adoerr (approved):
Thanks!

PR #253, opened 2021-07-28
Introduce test network and support components

adoerr:
This PR adds various building blocks in order to write unit tests which require either a network, client or backend.

To this end, the following structures have benn added to `beefy-test`
- `Client`
- `Peer`
- `Network`

The above components in it's current form should be considered a starting point. They provide what is needed to get going and will be enhanced / adapted based on the needs of concrete unit test cases.

### DIFF
--- a/beefy-test/src/lib.rs
+++ b/beefy-test/src/lib.rs
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+mod client;
 mod keyring;
 
+pub use client::Client;
 pub use keyring::Keyring;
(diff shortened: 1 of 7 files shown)


beefy-test/src/lib.rs
```diff
@@ -14,6 +14,13 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+mod client;
```
tomusdrw, line 17:
```suggestion
#![warn(missing_docs)]

mod client;
```

I'd suggest to add docs to all public items for better Developer Experience.

tomusdrw (approved):

PR #263, opened 2021-08-10
Use `tokio` for network tests

adoerr:
Use `tokio` for `beefy-test`


svyatonik (approved):
(I do not know the exact reasoning this change, but change itself is ok)

PR #269, opened 2021-08-18
Test finality notification

adoerr:


beefy-test/src/client_tests.rs
```diff
@@ -0,0 +1,147 @@
+// Copyright (C) 2021 Parity Technologies (UK) Ltd.
```
hbulgarini, line 1: is this the exact same test that it used to run from `client.rs` but now it runs on this dedicated test file?

adoerr: Yep, moved the unit tests into a separate file.

hbulgarini (approved):
lgtm!

PR #270, opened 2021-08-19
Make Peer::Client an Arc

adoerr:
- Bump Substrate and Deps
- Make `beefy-test::Peer::Client` and Arc


hbulgarini (approved):
I have read the PR a few times and i think that i get the scope of this PR that aims on the replacement of NativeExecutor by NativeElseWasmExecutor. 

However i cannot asses if this change might have an impact since my background is a little bit limited. If you consider this is good to go i rely on your experience you can go ahead with the change from my side.

PR #283, opened 2021-09-05
Update cargo fmt config

adoerr:
Use `nightly` and `imports_granularity="Crate"`


adoerr:
Will be addressed later.

PR #288, opened 2021-09-09
sp-io update

chevdor:
`cargo update sp-io`

The `polkadot-v0.9.10` in Substrate saw some changes so we need to update accordingly.


chevdor:
Closed because submitted against master.

PR #285, opened 2021-09-06
chore: update beefy to substrate branch polkadot-v0.9.10

chevdor:
This PR updates the substrate dependencies to the `polkadot-v0.9.10` branch.

See below for details about the 2 CI checks that are failing.

PR #289, opened 2021-09-09
sp-io update

chevdor:
This PR applies `cargo update sp-io` into the `polkadot-v0.9.10`.

The `polkadot-v0.9.10` in Substrate saw some changes so we need to update accordingly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,56 +38,37 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aead"
-version = "0.3.2"
+version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7fc95d1bdb8e6666b2b217308eeeb09f2d6728d104be3e31916cc74d15420331"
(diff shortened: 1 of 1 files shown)


chevdor:
@tomusdrw we will likely get in this PR the same CI failures as in #285 (I already see the `deny` issue). Do you want to handle those differently this for this PR?

adoerr (commented):
@chevdor I'm a bit confused, why the `sp-io` update is needed at the `BEEFY` level.

Whenver I updated the `BEEFY` dep in Polkadot I first do a
`cargo update -p beefy-primitves`
and then a
`cargo update  -p sp-io --precise [exact commit needed for Polkadot]`

This usaully gives the the minimal set of changes for `Cargo.lock`. What am I missing here :thinking:

chevdor:
The issue is that:
- beefy depends on Substrate
- diener nicely switched all the deps to the new Substrate branch `polkadot-v0.9.10`
- in the meantime, the branch evolved but the referenced commits did not

At the Polkadot level, running a `cargo update -p sp-io` gives my missmatch on some deps compared to `cargo update -p beefy-primitives`. So with this PR, I am trying to align the two.

Do you see any potential issue with this update ?

adoerr (approved):

PR #293, opened 2021-09-16
Change CI from Github Actions to Gitlab CI

alvicsam:
Confige .gitlab-ci.yml instead of GHA.

https://github.com/paritytech/ci_cd/issues/178


.gitlab-ci.yml
```diff
@@ -0,0 +1,107 @@
...
+  <<:                              *docker-env
+  script:
+    # add clippy is temporary until tests in substrate are not fixed
+    # https://github.com/paritytech/substrate/pull/9637
+    - rustup component add clippy --toolchain nightly-x86_64-unknown-linux-gnu
+    - cargo +nightly clippy --all-targets -- -D warnings
+
+#### stage:                       test
+
+deny:
```
tomusdrw, line 77: In previous config deny was also running periodically (nightly) to make sure changes are detected before a PR is created. Obviously it only makes sense with email notifications to detect the failures. If it's cumbersome to implement, we can keep it as-is.

.gitlab-ci.yml
```diff
@@ -0,0 +1,107 @@
...
+  stage:                           test
+  <<:                              *docker-env
+  script:
+    - cargo test --all  --verbose
+
+check:
+  stage:                           test
+  <<:                              *docker-env
+  script:
+    - cargo check --all  --verbose
```
tomusdrw, line 98: Do I get this right that `cargo check` will be performed after `cargo test`? I don't think it makes much sense - `check` is more lightweight (not doing compilation, just static analysis) and I think there isn't anything that would slip through `test` run.

tomusdrw (approved):

.gitlab-ci.yml
alvicsam: Not exactly: jobs `deny`, `test` and `check` are executed on the stage `test` which means that they run simultaneously on different runners. But I can divide it to different stages, just not sure which way is the best.

.gitlab-ci.yml
alvicsam: I think it's doable but currently it's still unclear for me whether I need to finish this PR or just close it since the repo is going to merged in Substrate project.

tomusdrw: Yeah, fair enough, let's skip it.

.gitlab-ci.yml
tomusdrw: Running in parallel sounds good, we can keep it as-is.

.gitlab-ci.yml
```diff
@@ -0,0 +1,107 @@
...
+  image:                           "${CI_IMAGE}"
+  <<:                              *kubernetes-env
+  script:
+    - cargo +nightly fmt --all -- --check --config merge_imports=true
+
+clippy:
+  stage:                           lint
+  <<:                              *docker-env
+  script:
+    # add clippy is temporary until tests in substrate are not fixed
```
TriplEight, line 70: better to prepend such lines with
```suggestion
    # FIXME: add clippy is temporary until tests in substrate are not fixed
```

.gitlab-ci.yml
```diff
@@ -0,0 +1,107 @@
...
+  - lint
+  - test
+  - build
+
+variables:                         &default-vars
+  GIT_STRATEGY:                    fetch
+  GIT_DEPTH:                       100
+  CARGO_INCREMENTAL:               0
+  CARGO_TARGET_DIR:                "/ci-cache/${CI_PROJECT_NAME}/targets/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}"
+  CI_IMAGE:                        "paritytech/ci-linux:production"
```
TriplEight, line 12: what image is better here actually?
https://github.com/paritytech/scripts/blob/master/dockerfiles/ci-linux/Dockerfile or
https://github.com/paritytech/scripts/blob/master/dockerfiles/bridges-ci/Dockerfile
@tomusdrw maybe you can tell?
Or is it just a question of compatibility with Substrate?

.gitlab-ci.yml
```diff
@@ -0,0 +1,107 @@
...
+    - cargo check --all  --verbose
+
+#### stage:                       build
+
+build:
+  stage:                           build
+  <<:                              *docker-env
+  script:
+    - *rust-info-script
+    - cargo build --all --release --verbose
```
TriplEight, line 107:
```suggestion
    - cargo build --all --release --verbose

```

TriplEight (changes requested):

adoerr (approved):

PR #320, opened 2021-10-21
BEEFY High-level docs

tomusdrw:
[Rendered](https://github.com/paritytech/grandpa-bridge-gadget/blob/td-docs/docs/beefy.md)

### DIFF
--- a/docs/beefy.md
+++ b/docs/beefy.md
@@ -92,11 +92,11 @@ A pseudo-algorithm of behavior for a fully-synced BEEFY validator is:
 
 ```
 loop {
-  let (best_beefy_block, best_grandpa_block) = wait_for_best_blocks();
+  let (best_beefy, best_grandpa) = wait_for_best_blocks();
 
   let block_to_vote_on = choose_next_beefy_block(
-    best_beefy_block,
-    best_grandpa_block
+    best_beefy,
+    best_grandpa
   );
 
   let payload_to_vote_on = retrieve_payload(block_to_vote_on);
@@ -299,9 +299,14 @@ Misbehavior should be penalized. If more validators misbehave in the exact same 
 should be more severe, up to the entire bonded stake in case we reach `1/3rd + 1` validators
 misbehaving.
 
+<details>
+<summary>Open questions</summary>
+
 // TODO [ToDr] Penalization formula.
 // TODO [ToDr] If there are misbehaving validators in GRANDPA should we penalize them twice?
 IMHO yes, cause they should be relying on GRANDPA finality proof, not just the votes.
+
+</details>
 
 # Implementation
 
@@ -340,8 +345,40 @@ TODO
 
 ## Lean BEEFY
 
-// TODO [ToDr] Mandatory justifications are helping out with gadget<>sync comms, but they are abit
-of a hack.
+To simplify the initial implementation of BEEFY, we are taking a bunch of shortcuts. These
+shotcuts are described in this section and are collectively called "Lean BEEFY".
+
+Lean BEEFY is based on additional assumption of validators always being able to complete the
+mandatory round, i.e. there is at least `2/3rd + 1` validators on-line during the session long
+enough to conclude at least one round. Based on this assumption, we further relax some requirements
+to simplify the implementation.
+
+### Justifications Sync
+
+The nodes are not required to sync BEEFY justifications in order, nor they are expected to
+have all justifications for mandatory blocks. Since the assumption is that these mandatory rounds
+have been concluded the nodes assume that justifications are available somewhere, and since BEEFY
+piggy-backs on GRANDPA, it's further assumed that since GRANDPA finalized blocks, BEEFY had to
+correctly finalize the same blocks as well.
+
+Implementation wise, this means that it's not required to plug into the block import pipeline or
+alter any `sync` logic. The node can only import justifications that it sees being gossiped when it's
+on-line. The complete sync of justifications for mandatory blocks can be implemented later.
+
+### Round Selection
+
+Since the node does not necessarily have justifications for mandatory blocks, when it sees GRANDPA
+moving to a new session, it can assume BEEFY has moved to that new session as well. Hence the
+`best_beefy` block number can immediatelly be set to the mandatory block number of the previous
+session, despite the fact the node does not have the justification for that block yet.
+
+### Mandatory Justification Gossip
+
+Due to the lack of `sync` alterations, that would ensure all justifications for mandatory blocks
+are fetched and imported to the local database, the node, after going on-line will miss recent
+justifications. To help out with justification propagation mandatory justifications are gossiped
+on the global topic as described in the [Gossip](#3-Gossip) section. While they are not strictly
+necessary for neither BEEFY nor Lean BEEFY, practically this should make the consensus more robust.
 
 # BEEFY & MMR (Polkadot implementation)
 
@@ -363,24 +400,3 @@ TODO
 
 TODO
 
-
-# Assorted notes
-
-Only one round is always active and it's in either of the modes:
-
-Mandatory mode
- - force sync to request beefy justifcations for this block
- - listen to justification imports
- - vote
-
-Non-mandatory mode
- - listen to  justification imports
- - vote on a block according to power-of-two rule
-
-
-How to get the initial state?
-
-
-We gossip votes for the current round.
-
-We gossip justifications for some past rounds?


docs/beefy.md
```diff
@@ -4,97 +4,364 @@ Related issues:
...
+Regular nodes are expected to:
+1. Receive & validate votes for the current round and broadcast them to their peers.
+1. Receive & validate BEEFY Justifications and broadcast them to their peers.
+1. Return BEEFY Justifications for [**Mandatory Blocks**](#3-round-selection) on demand.
+1. Optionally return BEEFY Justifications for non-mandatory blocks on demand.
+
+Validators are expected to additionally:
+1. Produce & broadcast vote for the current round.
+
+Both kinds of actors are expected to fully participate in the protocol ONLY IF they believe they
```
acatangiu, line 145: So until they catch up they are fully passive?

docs/beefy.md
```diff
@@ -4,97 +4,364 @@ Related issues:
...
+As mentioned earlier, every time the node picks a new `round_number` (and validator casts a vote) it
+ends the previous one, no matter if finality was reached (i.e. the round concluded) or not. Votes
+for an inactive round should not be propagated.
+
+Note that since BEEFY only votes for GRANDPA-finalized blocks, `session_start` here actually means:
+"the latest session for which the start of is GRANDPA-finalized", i.e. block production might have
+already progressed, but BEEFY needs to first finalize the mandatory block of the older session. See
+[Catch up](#3-catch-up) for more details and also consult [Lean BEEFY](#2-lean-beefy).
+
+In a good networking conditions BEEFY may end up finalizing each and every block (if GRANDPA does
```
acatangiu, line 214: nit:
```suggestion
In good networking conditions BEEFY may end up finalizing each and every block (if GRANDPA does
```

docs/beefy.md
```diff
@@ -4,97 +4,364 @@ Related issues:
...
+the same). Practically, with short block times, it's going to be rare and might be excessive, so
+it's suggested for implementations to introduce a `min_delta` parameter which will limit the amount
+of started rounds. The affected component of the formula would be: `best_beefy + MAX(min_delta,
```
acatangiu, line 217: Initially I read this as multiple rounds being active at the same time, I suggest a reword for clarity:

```suggestion
it's suggested for implementations to introduce a `min_delta` parameter which will limit the frequency with which new rounds are started. The affected component of the formula would be: `best_beefy + MAX(min_delta,
```

docs/beefy.md
```diff
@@ -4,97 +4,364 @@ Related issues:
...
+### Initial Sync
+
+It's all rainbows and unicorns when the node is fully synced with the network. However during cold
+startup it will have hard time determining the current round number. Because of that nodes that
+are not fully synced should not participate in BEEFY protocol at all.
+
+During the sync we should make sure to also fetch BEEFY justifications for all mandatory blocks.
+This can happen asynchronously, but validators, before starting to vote, need to be certain
+about the last session that contains a concluded round on mandatory block in order to initiate the
+catch up procedure.
```
acatangiu, line 250: How does this happen?
Do nodes/validators just passively listen for votes for a while until they know about the last session that contains a concluded round on mandatory block?

Do we have a plan for handling > 1/3 validators in this state?

docs/beefy.md
```diff
@@ -4,97 +4,364 @@ Related issues:
...
+conclude rounds for mandatory blocks. Note that older sessions must obviously be finalized by the
+validator set at that point in time, not the latest/current one.
+
+### Initial Sync
+
+It's all rainbows and unicorns when the node is fully synced with the network. However during cold
+startup it will have hard time determining the current round number. Because of that nodes that
+are not fully synced should not participate in BEEFY protocol at all.
+
+During the sync we should make sure to also fetch BEEFY justifications for all mandatory blocks.
```
acatangiu, line 247: Why do we need justifications for all mandatory blocks? why not just the last mandatory from the current/previous session?

docs/beefy.md
```diff
@@ -4,97 +4,364 @@ Related issues:
...
+
+## Misbehavior
+
+Similarily to other PoS protocols, BEEFY considers casting two different votes in the same round a
+misbehavior. I.e. for a particular `round_number`, the validator produces signatures for 2 different
+`Commitment`s and broadcasts them. This is called **equivocation**.
+
+On top of this, voting on an incorrect **payload** is considered a misbehavior as well, since
+because we piggy-back on GRANDPA there is no ambiguity in terms of the fork validators should be
+voting for.
```
acatangiu, line 291:
```suggestion
On top of this, voting on an incorrect **payload** is considered a misbehavior as well, and since we piggy-back on GRANDPA there is no ambiguity in terms of the fork that validators should be
voting for.
```

docs/beefy.md
```diff
@@ -4,97 +4,364 @@ Related issues:
...
+
+Misbehavior should be penalized. If more validators misbehave in the exact same `round` the penalty
+should be more severe, up to the entire bonded stake in case we reach `2/3rd + 1` misbehaving
```
acatangiu, line 294: should this be `1/3`? Can we handle more than `1/3` dishonest validators?

docs/beefy.md
```diff
@@ -4,97 +4,364 @@ Related issues:
...
+Nodes participating in BEEFY protocol are expected to gossip messages around. The protocol defines
+following messages:
+
+1. Votes for the current round,
+2. BEEFY Justifications for recently concluded rounds,
+3. BEEFY Justification for the latest mandatory block,
+
+Each message is additionally associated with a **topic**, which can be either:
+1. the round number (i.e. topic associated with a particular round),
+2. or the global topic (independent from the rounds).
```
acatangiu, line 263: So to be clear, `votes for the current round` (1) use a _per-round specific topic_, while justifications (2&3) use the global topic?

docs/beefy.md
```diff
@@ -4,97 +4,378 @@ Related issues:
...
+    - The commitment matches local commitment.
+    - The validator is part of the current validator set.
+    - The signature is correct.
+
+- **BEEFY Justification**
+  - Justifications are sent on the global topic.
+  - Justification is considered valid when:
+    - It is for a recent (implementation specific) round or the latest mandatory round.
+    - It has at least `2/3rd + 1` valid signatures.
+    - Signatorees are part of the current validator set.
```
acatangiu, line 280:
```suggestion
    - Signatorees are part of the validator set pertaining to the active session in respective round.
```

docs/beefy.md
```diff
@@ -4,97 +4,378 @@ Related issues:
...
+  - Votes are sent on the round-specific topic.
+  - Vote is considered valid when:
+    - The commitment matches local commitment.
+    - The validator is part of the current validator set.
+    - The signature is correct.
+
+- **BEEFY Justification**
+  - Justifications are sent on the global topic.
+  - Justification is considered valid when:
+    - It is for a recent (implementation specific) round or the latest mandatory round.
```
acatangiu, line 278:
```suggestion
    - It is for a recent (implementation specific) round or a mandatory round.
```
?

docs/beefy.md
```diff
@@ -4,97 +4,364 @@ Related issues:
...
+using block number instead of a hash. A collection of **votes**, or rather
+a Commitment and a collection of signatures is going to be called **Signed Commitment**. A valid
+(see later for the rules) Signed Commitment is also called a **BEEFY Justification** or
+**BEEFY Finality Proof**. For more details on the actual data structures please see
+[Data Formats](#2-data-formats).
+
+A **round** is an attempt by BEEFY validators to produce BEEFY Justification. **Round number** is
+simply defined as a block number the validators are voting for, or to be more precise, the
+Commitment for that block number. Round ends when the next round is started or when we receive ALL
+expected votes from ALL validators.
```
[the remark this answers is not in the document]
tomusdrw: Hmm, good point. We should start the new round as soon as we get 2/3+1 votes, so the second part is redundant.

I've rephrased the whole sentence and added more clarity.

docs/beefy.md
tomusdrw: Correct. Added an extra sentence to emphasize that.

docs/beefy.md
tomusdrw: We need to make sure that the validator sets are properly changing, so that BEEFY is kind of self-sufficient, i.e. each BEEFY validator set signs-off it's successor. Without that you would just jump in the middle and assume that the current validator set is indeed legit. While this should be ensured by GRANDPA, we should rather make sure that we don't need this assumption  and also be sure that each full node can serve any BEEFY light client (i.e. contains all mandatory justifications).

docs/beefy.md
tomusdrw: No, a node should actively request BEEFY justifications for mandatory blocks from it's peers until it has all of them (i.e. full BEEFY-signed history of all validator set transitions).

If we have > 1/3 valdiators in that state, BEEFY will simply start falling back behind GRANDPA - i.e. the validators will be stuck at some mandatory round, not being able to collect enough votes, but when 2/3 + 1 get's fully synced (and also stuck on the same round) we should start seeing votes being produced, gossiped and eventually the round being concluded.

docs/beefy.md
tomusdrw: Yes. It's described in the per-message section below.

docs/beefy.md
tomusdrw: Well, here I've rather meant "valid" in terms of "worthwhile to gossip", indeed we can have some old justification that is valid, but we shouldn't be gossiping it (hence "recent round" and "the latest mandatory round"). I've rephrased in the doc as well.

docs/beefy.md
tomusdrw: Again, since this section defines the gossip, I'd rather narrow this down to usefulness in the latest view of the BEEFY state.

docs/beefy.md
tomusdrw: Yes, you're right.

docs/beefy.md
```diff
@@ -4,97 +4,399 @@ Related issues:
...
+   (i.e. we are doing BEEFY Catch-up)
+4. or if we can follow the power-of-two rule.
+
+With LEAN BEEFY:
+1. Start with Best GRANDPA block
+2. Traverse up to "Session Length" number of blocks backwards to find the session start block.
+3. ASSUME this block is BEEFY-finalized, even though we haven't seen BEEFY Justification.
+
+
+//TODO [ToDr] How to figure out the `session_start` block?
```
acatangiu, line 334: PR https://github.com/paritytech/substrate/pull/10669 implement `current_session_start_block_num` storage for session pallet. That should work fine, right?

acatangiu (approved):
Let's merge this and continue work on focused topics in following PRs

PR #365, opened 2021-12-07
mmr-cli: verify MmrLeaf version against Rococo

### LINKED ISSUE #255: Verify MMRLeaf version
Currently the `decode-leaf` subcommand will happily decode every MMRLeaf - instead we should make sure to trigger a warning if the decode version does not match the expected one to avoid decoding BS.

acatangiu:
Once Beefy is integrated to Polkadot, the cli should verify against Polkadot leaf version.

Fixes #255

### DIFF
--- a/beefy-cli/src/cli/mmr.rs
+++ b/beefy-cli/src/cli/mmr.rs
@@ -15,10 +15,15 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 use crate::cli::utils::Bytes;
-use beefy_primitives::mmr::MmrLeaf;
+use beefy_primitives::mmr::{MmrLeaf, MmrLeafVersion};
 use parity_scale_codec::{Decode, Encode};
 use sp_core::H256;
 use structopt::StructOpt;
+
+// Hardcoded leaf version from Rococo/Polkadot runtime.
+fn polkadot_leaf_version() -> MmrLeafVersion {
+	MmrLeafVersion::new(0, 0)
+}
 
 /// MMR related commands
 #[derive(StructOpt)]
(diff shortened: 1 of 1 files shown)


deny.toml
```diff
@@ -69,7 +69,7 @@ ignore = [
 # https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html
 [licenses]
 # The lint level for crates which do not have a detectable license
-unlicensed = "deny"
+unlicensed = "allow"
```
tomusdrw, line 72: I'd rather keep the lint and whitelist a package if it's really needed, is that possible?

beefy-cli/Cargo.toml
```diff
@@ -18,6 +18,9 @@ structopt = "0.3.21"
 beefy-primitives = { git = "https://github.com/paritytech/substrate", branch = "master" }
 beefy-merkle-tree = { git = "https://github.com/paritytech/substrate", branch = "master" }
 
+# Polkadot dependencies
+rococo-runtime = { git = "https://github.com/paritytech/polkadot", branch = "master" }
```
tomusdrw, line 22: Let's not depend on the entire polkadot repo just for this. Can you please just hardcode that? The leaf data structure is hardcoded as well anyway (coming from substrate primitives), it's not like we will be automatically able to support newer versions.

beefy-cli/src/cli/mmr.rs
```diff
@@ -55,6 +56,13 @@ impl Mmr {
 				};
 				let leaf: Vec<u8> = Decode::decode(&mut leaf_content)?;
 				let leaf: MmrLeaf<u32, H256, H256> = Decode::decode(&mut &*leaf)?;
+				let expected_leaf_vers = LeafVersion::get();
+				if leaf.version != expected_leaf_vers {
```
tomusdrw, line 60: I'm wondering if it makes sense to compare the version components separately and refuse even printing the leaf if the major part does not match (that would require partially decoding the leaf first though). but perhaps it's an overkill.

beefy-cli/Cargo.toml
acatangiu: done, ptal

tomusdrw (approved):

